PR #1, opened 2026-02-27
docs: robot interface proposal

maxxgx:
```python
class Robot(Protocol):
    def connect(self) -> None: ...
    def disconnect(self) -> None: ...
    def get_observation(self) -> dict[str, Any]: ...
    def send_action(self, action: np.ndarray) -> None: ...
```


docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+
+The Protocol cannot provide default method implementations. Instead, `physicalai` ships a context manager wrapper:
+
+```python
+# physicalai/robots/utils.py
+from contextlib import contextmanager
+
+
+@contextmanager
+def managed(robot):
```
samet-akcay: I guess this is where protocol falls short.

it might be an ok compromise but not sure about the name. maybe could think about alternative names ?

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+with managed(robot):
+    obs = robot.get_observation()
+    action = model(obs)
+    robot.send_action(action)
+```
+
+## Design Principles
+
+- **Protocol, not inheritance.** No base class to subclass. Implementations are plain classes that happen to have the right methods.
+- **Standard Python types.** Observations are `dict[str, Any]`, actions are `np.ndarray`. No custom message types, no protobuf, no ROS messages.
```
MarkRedeman: Would it be possible provide this type as a generic? That way we might be able to use more strict type inference.

I.e. suppose you have,
```python
lekiwi = LeKiwi(lekiwi_config)
so101_leader = SO101(so101_config)

leader_observarion = so101_leader.get_observation() # as `dict[str, Any]`
lekiwi.send_action(leader_observation) # throws because lekiwi prepends `arm_` to the SO101 arm's joint names
```

If we were to be able to add generic types (as an opt in, shouldn't be strictly required as it can also be annoying), then we might have,
```python
def reformat_names(observation: SO101Observation):
    return  { 'arm_shoulder_pan.pos': observation['shoulder_pan.pos'], ... }

leader_observarion = so101_leader.get_observation() # as `{ 'shoulder_pan.pos': int`, ... }
lekiwi.send_action(reformat_names(leader_observation))
```

the second example will be a lot more verbose, but prevents you from applying the wrong actions.

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+
+    def disconnect(self) -> None:
+        # stop motors, close connection
+        ...
+
+    def get_observation(self) -> dict[str, Any]:
+        return {
+            "state": np.array([...]),
+            "timestamp": time.monotonic(),
+        }
```
MarkRedeman: Note on this structure: I added this structure initially to the Geti Action repo so that we could use `timestamp` as a way to do time synchronization for robot states and cameras.

I think that if we were to add this as examples then we should try to make this a strict part of the protocol. So that we'd have `get_observation(self) -> { state: dict[str, Any], timestamp: int }` (not sure if this is the correct python syntax, but something like that..)

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+            self._connection.close()
+            self._connection = None
+        for cam in self.cameras.values():
+            cam.stop()
+
+    def get_observation(self) -> dict[str, Any]:
+        images = {
+            name: cam.read().data
+            for name, cam in self.cameras.items()
+        }
```
MarkRedeman: I'd personally not make cameras be part of a robot interface / implementation. 
Lerobot does this, but it complicates connecting to robots as you'd need to make sure the cameras are also connected, even if you only care about reading the joint states.

Another reason for not doing coupling them is that you get ambiguity when a camera might be used by two robots. E.g. for aloha mini you can consider its two SO101 arms as two individual robots, which share the same "overview" cameras.

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+            "timestamp": time.monotonic(),
+        }
+
+    def send_action(self, action: np.ndarray) -> None:
+        left_action = action[:6]
+        right_action = action[6:]
+        ...
+```
+
+**Multiple independent robots**: Multiple instances, managed separately.
```
MarkRedeman: Personally I quite like this approach as it feels simpler.
One thing to consider though is that when storing the joint states etc of a robot, we'd need to make sure their feature names are prepended with say `left_`, `right_`, or use some other "merging strategy".

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+    # obs["images"]["wrist"] contains the camera frame
+```
+
+### User-managed cameras
+
+User reads cameras separately and assembles the observation dict:
+
+```python
+camera = OpenCVCamera(index=0)
+robot = SO100(port="/dev/ttyUSB0")
```
MarkRedeman: In Geti Action we add one more abstraction on top of this,
```python
environment_setup = Evnrionment(
  robots = [so100(port="/dev/ttyUSB0")],
  cameras = [OpenCVCamera(index=0)]
)
```

The main thing this solves is that we don't have ambiguity on cameras belonging to a specific robot.
Another thing I have on my mind is that we could have methods like `environment_setup.is_compatible(dataset_1)`,`environment_setup.is_compatible(model_policy)`, `environment_setup.is_compatible(another_environment)`  etc.
An environment might also be able to produce `environment_setup.manifest()` which could be used as a sort of record to keep track of how datasets are recorded etc.

I don't think we'd need to consider this for this interface proposal, but figured it might be worth to share these concepts.

docs/design/components/proposal-robot-interface.md
MarkRedeman: One other reason for keeping cameras and robots separate: by allowing users to do `robot.get_observation()` and `camera.get_observation()` (or `camera.read()` etc), we can allow the user to record the robot's state and camera state with different frequency / fps.

While this is hard to do with LeRobot's datasets (as its optimized for training) it's not too hard to allow this when we write our own custom dataset adapter.

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+            f"Loop overrun: {elapsed:.3f}s > {target_dt:.3f}s. "
+            f"Inference may be too slow for the target frequency."
+        )
+```
+
+When inference is slower than the target frequency, action chunking runners can compensate by executing multiple pre-computed actions between inference calls. This is handled at the runner level, not the robot level.
+
+## Async and Concurrency
+
+The Protocol is synchronous. All methods block until complete.
```
MarkRedeman: One thing to consider: the [trossen api](https://docs.trossenrobotics.com/trossen_arm/main/api/classtrossen__arm_1_1TrossenArmDriver.html#_CPPv4N11trossen_arm16TrossenArmDriver17set_arm_positionsERKNSt6vectorIdEEdbRKNSt8optionalINSt6vectorIdEEEERKNSt8optionalINSt6vectorIdEEEE) allows users to set a `goal_time` and also a `blocking` argument.

I think having helper classes that allow this could be useful for us.

docs/design/components/proposal-robot-interface.md
```diff
@@ -0,0 +1,608 @@
...
+    "io": {
+        "inputs": {
+            "images.wrist": { "shape": [3, 480, 640], "dtype": "uint8" },
+            "state": { "shape": [6], "dtype": "float32" }
+        },
+        "outputs": {
+            "action": { "shape": [100, 6], "dtype": "float32" }
+        }
+    }
+}
```
maxxgx: I think the content of the manifest is still unclear:
- Will the manifest contain info about the robot it was trained on?
- Is a fine-tuned policy only compatible with a single environment (robot + camera setup)?

Update: we agreed to have robot and camera config in the manifest file

docs/design/components/proposal-robot-interface.md
maxxgx: Maybe we can be more explicit: `with robot_connection(robot)`.
Or maybe `with connect(robot)` (like open for files)?

edit: I propose to go with "connect"

docs/design/components/proposal-robot-interface.md
maxxgx: We should be able to enforce this structure using `TypedDict`. However, I'd wait before doing so, we can start being less strict with `dict[str, Any]` and change it to a `TypedDict` later on if needed.

docs/design/components/proposal-robot-interface.md
maxxgx: Thanks for pointing this out. I agree that the robot should not own the cameras. The pairing robot + camera should be managed by the orchestrator (e.g. `physicalai.runtime`)

docs/design/components/proposal-robot-interface.md
maxxgx: It's not very clear what happens with trossen's api when `blocking=False`. What other actions can be taken when run async?

MarkRedeman: So from what I understood if `blocking=False` the call will send the request to the arm, but not wait for confirmation that the arm has moved (calling `set_arm_position` requires a network request as the arm's aren't connected over a serial bus). 
I haven't been able to test this myself though as I don't have one of those arms, maybe Arend Jan can provide some more input here.

docs/design/components/proposal-robot-interface.md
maxxgx: While I agree with that robot and camera should ideally be separate. The interface is deliberately open, so if the user wants to combine robot + camera, we cannot stop them. Though I can update the example to not include the camera

docs/design/components/proposal-robot-interface.md
maxxgx: `set_arm_position` appears to be a synch method, so I guess it can only run async internally within the function. If so, having only a synchronous protocol should be fine

docs/design/components/proposal-robot-interface.md
maxxgx: Do you mean something like this?
```
T = TypeVar("T")
get_observation() ->  dict[str, T]
```

With the above, I doubt there are any benefit as if you have one of value in the dict with a different type, then it will be essentially same as `dict[str, Any]`

docs/design/components/robot-interface.md
```diff
@@ -1,720 +1,741 @@
...
+
+with connect(robot):
+    robot_obs = robot.get_observation()
+    frame = camera.read()
+
+    obs = {
+        "images": {"wrist": frame.data},
+        "state": robot_obs["state"],
+        "timestamp": robot_obs["timestamp"],
+    }
```
MarkRedeman, line 45: Possibly a nit pick here: the timestamp of a robot's state and the camera frames might not be the same.
I don't think this is too important for Geti Action atm since we only allow 30fps recordings atm. But if we were to allow to record the robot state at a higher frequency then we may need to synhronize them.

docs/design/components/robot-interface.md
```diff
@@ -1,720 +1,741 @@
...
+### Validation via Manifest
 
-This ensures the base interface remains portable while providing convenience for adapter packages.
+The robot does not describe itself. The policy's `manifest.json` declares what it expects. The runtime validates observations against the manifest on first contact. See the [Validation](#validation) section below.
 
-### Multi-Robot vs Multi-Arm
+### Safe Disconnect
 
-This design distinguishes **multiple robots** from **multi-arm robots**:
+`disconnect()` must leave the robot in a safe, stationary state. Motors must be stopped or holding position before the connection is closed. This is a contractual requirement on every implementation, verified by the [conformance test suite](#conformance-testing).
```
MarkRedeman, line 104: Does this also mean that, when we remove the torque on motors, we move the robot back to its home position when disconnecting?
I believe that Trossen has support for this as well. @ArendJanKramer

samet-akcay (approved):

docs/design/components/robot-interface.md
```diff
@@ -1,720 +1,741 @@
...
-        *,
-        id: str = "robot",
-        cameras: dict[str, dict[str, Any]] | None = None,
-        **kwargs: Any,
-    ) -> None:
-        """Initialize vendor robot wrapper.
+    def disconnect(self) -> None:
+        # Stop all motors before closing connection
+        if self._connection:
+            self._write_torque(enabled=False)
```
MarkRedeman: This contradicts the previous message that the robot should be in a stable state: disabling the torque could make it so that the arm suddenly drops due to gravity.

docs/design/components/robot-interface.md
```diff
@@ -1,720 +1,741 @@
...
+```json
+{
+    "format": "policy_package",
+    "version": "1.0",
+    "robots": [
+        {
+            "name": "main",
+            "type": "SO-100",
+            "state": { "shape": [6], "dtype": "float32" },
+            "action": { "shape": [6], "dtype": "float32" }
```
MarkRedeman: I thought that in lerobot they expect a dict like,
```
{
            'shoulder_pan.pos': -8.705526116578355,
            'shoulder_lift.pos': -98.16753926701571,
            'elbow_flex.pos': 95.98393574297188,
            'wrist_flex.pos': 73.85993485342019,
            'wrist_roll.pos': -13.84615384615384,
            'gripper.pos': 26.885644768856448
}
```

vs an vector,
```
[
    -8.705526116578355,
    -98.16753926701571,
    95.98393574297188,
    73.85993485342019,
    -13.84615384615384,
    26.885644768856448
]
```

I guess having it as a vector makes it easier: for bimanual robots we can concatenate the vectors and don't have to deal with adding a string prefix.
Main con would be that it's a bit more fault tolerant (e.g. do you `np.concatenate([left_joints, right_joints])`, or `np.concatenate([right_joints, left_joints])`?).

To me both would be fine, but figured it would be good to call out this difference.

docs/design/components/robot-interface.md
maxxgx: Not necessarily, the only requirement is that the robot should not move after it's been disconnected, as that could be dangerous. The implementation can decide whether to go back to a predefined position or freeze

docs/design/components/robot-interface.md
maxxgx: Nice catch, I updated the example to keep torque enabled

docs/design/components/robot-interface.md
maxxgx: I've added `joint_order` in the manifest to make the contract clearer. However, I think it should be optional since third parties shouldn't be forced to provide it

docs/design/components/robot-interface.md
```diff
@@ -1,720 +1,778 @@
...
-### Multi-Arm Schema Example
+The model's `manifest.json` provides robot type, state/action shapes, and camera requirements. The user supplies hardware connection details via `--port`/`--device` flags or a local config file (flags override config):
 
-Example schema for a bimanual robot with 7‑DoF per arm:
+```bash
+# Port flag
+physical-ai infer \
+    --model ./exports/act_policy \
+    --port /dev/ttyUSB0 \
+    --episodes 10
```
MarkRedeman, line 627: I'm not really sure how `--episodes` would apply here. I'd assume this is something that would belong to an application side an not necessarily on a library.

Though let's revisit that at a later point in time if needed, I guess this interface will converge a bit more when we have finished the implementation and integration?

MarkRedeman (approved):

docs/design/components/robot-interface.md
MarkRedeman: Ok that sounds good to me. Perhaps later on we can extend with an extra protocol like,
```
class HasHomePosition(Protocol)
    def home_position(self) -> np.array: ...
```

where this would then return the "stable home position", i.e. for a SO101 robot that would roughly be,
```
{
    "shoulder_pan.pos": 0.0,
    "shoulder_lift.pos": -90,
    "elbow_flex.pos": 100.0,
    "wrist_flex.pos": 75.0,
    "wrist_roll.pos": 100.0,
    "gripper.pos": 0.0
}
```
these are the angles in degrees where the SO101 looks kind of folded, so it won't suddenly drop down if the motors loose their torque. So then the client could run `robot.send_action(robot.home_position())` to make it go into that home / rest position.

docs/design/components/robot-interface.md
maxxgx: I think the idea is to give the option in `physical-ai` to orchestrate a series of episodes/tasks and reset after each one. But yeah, we'll revisit it later on 😄

PR #2, opened 2026-03-06
docs: update camera interface design

maxxgx:
Main changes from the original design:

- **Flattened hierarchy**: Removed `Sensor` ABC, single `Camera` ABC now. Playback sources will use a `Protocol` later if needed.
- **`connect()` guarantees frames**: Blocks until first frame arrives or raises `CaptureTimeoutError`. Avoids startup race conditions.
- **Timeouts on `read()`**: Prevents indefinite blocking on hardware hangs.
- **Multi-camera sync**: Added `read_cameras()` / `async_read_cameras()` with skew reporting and `latest` flag for recording vs inference use cases.
- **Device discovery**: `DeviceInfo` now has `driver` + `hardware_id` for cross-backend dedup. Multiple instances on same device documented as undefined behavior.
- **Dropped `__iter__`**: Avoids silent data loss from error swallowing. Explicit `read()` is clearer.
- **Lazy async executor**: Per-camera thread pool only created if you actually use `async_read()`. Zero overhead for sync-only usage.
- **New sections**: Config validation, testing strategy, logging (loguru), color mode contracts.

### DIFF
--- a/docs/design/components/camera-interface.md
+++ b/docs/design/components/camera-interface.md
@@ -306,7 +306,13 @@ class Camera(ABC):
     @property
     @abstractmethod
     def device_id(self) -> str:
-        """Stable identifier for the physical device.
+        """Identifier for the physical device this instance targets.
+
+        Stable for the lifetime of the connection. May change across
+        reconnects for OS-assigned paths (e.g., /dev/video0).
+
+        Should match the corresponding DeviceInfo.device_id returned
+        by discover() for the same device.
 
         Examples: "/dev/video0", "serial:12345678", "rtsp://192.168.1.100/stream"
         """
--- a/docs/design/components/camera-interface.md
+++ b/docs/design/components/camera-interface.md
@@ -14,11 +14,12 @@
   - [Dependencies](#dependencies)
   - [Core Interface](#core-interface)
     - [Frame](#frame)
-    - [Sensor ABC](#sensor-abc)
     - [Camera ABC](#camera-abc)
+    - [Errors](#errors)
     - [DeviceInfo](#deviceinfo)
     - [Capability Mixins](#capability-mixins)
   - [Read Semantics](#read-semantics)
+    - [Why No Iterator Protocol](#why-no-iterator-protocol)
   - [Proposed Implementations](#proposed-implementations)
     - [OpenCVCamera](#opencvcamera)
     - [RealSenseCamera](#realsensecamera)
@@ -26,6 +27,7 @@
     - [GenicamCamera](#genicamcamera)
     - [IPCamera](#ipcamera)
   - [Recorded Sources (Future)](#recorded-sources-future)
+  - [Multi-Camera Synchronization](#multi-camera-synchronization)
   - [Factory Function](#factory-function)
   - [Sharing Model](#sharing-model)
   - [Usage](#usage)
@@ -40,6 +42,9 @@
     - [API Migration Map](#api-migration-map)
     - [Migration Plan](#migration-plan)
   - [Comparison with LeRobot](#comparison-with-lerobot)
+  - [Logging](#logging)
+  - [Testing Strategy](#testing-strategy)
+  - [Configuration Validation](#configuration-validation)
   - [Open Design Decisions](#open-design-decisions)
   - [References](#references)
 
@@ -56,9 +61,11 @@ This document defines the camera/capture interface for the physical-AI ecosystem
 - **Primary API**: Dedicated camera classes (`OpenCVCamera`, `RealSenseCamera`, etc.) with explicit constructor parameters
 - **Convenience API**: Thin `create_camera()` factory for config-driven workflows
 - **Read model**: Three-tier — `read()` (blocking sequential), `read_latest()` (non-blocking latest), `async_read()` (async/await)
-- **Frame type**: `Frame` dataclass carrying image data + timestamp + sequence number
+- **Multi-camera**: `read_cameras()` / `async_read_cameras()` for temporally aligned multi-camera reads
+- **Frame type**: `Frame` frozen slotted dataclass carrying image data + timestamp + sequence number
 - **Sharing**: Explicit only — no invisible global state, no hidden reference counting
-- **Callbacks**: Removed from base class — application layer concern
+- **Callbacks**: Removed from base class. Application layer concern
+- **No iterator protocol**: `Camera` does not implement `__iter__`. Explicit `read()` calls are clearer and avoid silent error swallowing
 
 **Goal**: The best camera framework for vision and robotics applications. Clean API, production-grade quality, hardware-agnostic.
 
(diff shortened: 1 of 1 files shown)


docs/design/components/camera-interface.md
```diff
@@ -129,25 +134,27 @@ from physicalai.capture import create_camera, discover_all
...
-Sensor (ABC)                       # Base: connect/disconnect/read/read_latest/async_read
-├── Camera (ABC)                   # Adds: device discovery, hardware config (width/height/fps)
-│   ├── OpenCVCamera               # USB webcams via OpenCV
-│   ├── RealSenseCamera            # Intel RealSense (+ DepthMixin)
-│   ├── BaslerCamera               # Basler industrial cameras (pypylon)
-│   ├── GenicamCamera              # Generic GenICam devices (harvesters)
-│   └── IPCamera                   # RTSP/HTTP network cameras
-└── [Future] RecordedSource        # Non-live playback
-    ├── VideoSource                # Video file playback
-    └── ImageDirectorySource       # Image sequence playback
```
maxxgx, line 141: Removing `Sensor ABC` for two reasons:
- Only Camera inherits it today. A base class with a single subclass is premature abstraction.
- The future subclasses don't fit the interface. VideoSource and ImageDirectorySource would inherit Sensor, but Sensor defines read_latest(), which does not make sense for recorded/playback sources.

docs/design/components/camera-interface.md
```diff
@@ -204,48 +211,134 @@ class Frame:
...
     @abstractmethod
     def is_connected(self) -> bool:
-        """Whether the source is currently open."""
+        """Whether the camera is currently open."""
+        ...
+
+    @property
+    @abstractmethod
+    def device_id(self) -> str:
+        """Stable identifier for the physical device.
```
MarkRedeman: For linux the identifiers are only stable as long as the device is connected, if you unplug and replug then changes are the identifier changed, e.g. `/dev/video0` to /dev/video2`.

I think for this design it's okay, but it's something we sadly need to deal with in the application code.

For the library perhaps one guarantee we could ask is that the identifier must be stable as long as the device is connected?

docs/design/components/camera-interface.md
MarkRedeman, line 207: Should `Frame` include an enum of the expected color mode?

docs/design/components/camera-interface.md
```diff
@@ -680,7 +778,80 @@ class ImageDirectorySource(Sensor):
...
+        timeout: Max seconds to wait for all cameras to respond.
+        latest: If True (default), uses read_latest() for freshest frame.
+            If False, uses read() for sequential capture. Use False for
+            recording workflows where every frame matters.
+
+    Raises:
+        CaptureTimeoutError: One or more cameras didn't respond in time.
+        CaptureError: A camera failed during read.
+    """
+    ...
```
MarkRedeman, line 832: I'm not sure what the intention is here, do we want to show the full implementation?

docs/design/components/camera-interface.md
```diff
@@ -943,49 +1168,110 @@ formats = camera.get_supported_formats()  # via FormatDiscoveryMixin
...
+responsibility. By default, nothing is printed.
 
-**3. Thread Model**
-One background capture thread per camera, or a shared thread pool? Per-camera threads are simpler but scale poorly with many cameras. Likely: per-camera threads initially, optimize later.
+---
 
-**4. Error Recovery**
+## Testing Strategy
+
+- **Unit tests**: A `FakeCamera(Camera)` subclass that returns pre-built frames is
```
MarkRedeman, line 1228: Can the pre build frames be set as a constructor argument? That way you could have two fake cameras one producing a blue frame and another a red frame. That might make testing easier as you can easily spot the differences between them.

docs/design/components/camera-interface.md
```diff
@@ -943,49 +1168,110 @@ formats = camera.get_supported_formats()  # via FormatDiscoveryMixin
...
+This best-effort approach avoids hard failures in config-driven workflows where
+parameters may come from a YAML file written for different hardware. Applications
+that require exact parameters should assert after `connect()`.
+
+---
+
+## Open Design Decisions
+
+**1. Multi-Consumer Access**
+If multiple consumers need the same camera, should the library provide a `CameraPool`? Or is passing the same instance sufficient? Current position: application layer concern. Revisit if multiple teams hit this need.
```
MarkRedeman, line 1270: This is one thing that I do like about the current framesource implementation: we don't need to be worried about our app breaking due to multiple users consuming the same camera feed. (my rust implementation on the other hand currently panics when two users try to read from the same camera source)

I do agree though that this likely is more of an application concern (same with having multiple consumers wanting to access the same robot).
There might also be two different behaviours you could have:
1. A pool that shares the same referecne to multiple consumers
2. A subscription that allow multiple consumers to subscribe to new frames, making it so that you can have multiple websocket handlers read from the same camera, while only doing the actual hardware read once. This gets tricky though if you'd allow to set custom fps per consumer.

(mainly sharing my thoughts here, no need to follow up on it atm)

MarkRedeman (approved):
I like the design and especially the explicit expected exceptions will make it a lot easier for us to do error handling on the application side.
The discovery and `FormatDiscoveryMixin` will be very useful for the UI integration.

The definition of `Frame` with its timestamp and sequence will help if we need to do data synchronization later on.

docs/design/components/camera-interface.md
maxxgx: I don't think that's necessary, as it should be clear what color mode you're using from the camera constructor argument

docs/design/components/camera-interface.md
maxxgx: I think the idea of this function should be clear. I left the implementation out since I wanted to focus on the interface (and not sure whether it would work 100% 😅).

But I believe we need to run each camera read inside their own threads for the sync version. It should be much simpler for the async version where we can use `asyncio.create_task` and gather the results after that.

docs/design/components/camera-interface.md
maxxgx: Yeah that's quite annoying, I think it also happens in MacOs. I've updated the docstring

docs/design/components/camera-interface.md
maxxgx: I think that should be possible to mock

docs/design/components/camera-interface.md
samet-akcay: I agree that `VideoSource` and `ImageDirectorySource` don't really fit here, and we might need to find a more elegant solution.

Regarding `Sensor`, multimodality will be on our roadmap in the future. IRD team already asked for it. A base `Sensor` could avoid repainting the architecture later. `Sensor` could perhaps be minimal and not camera-shaped. Maybe something like Sensor (generic): connect(), disconnect(), is_connected, read(timeout=...)

Thoughts?

maxxgx: I'm not convinced a `Sensor` base ABC is necessary for supporting multimodality. It would provide two things:

1. A shared read interface. But I don't see a use case for heterogeneous polymorphic reads:
```python
for sensor in sensor_list:
    sensor.read()  # What type does this return? Frame? Temperature? Sound? Proximity?
```
In practice, the policy expects a structured dict where each source is referenced by name and type:
```python
proximity_data = proximity.read()
robot_obs = robot.get_observation()

action = policy.select_action({
    "images": {
        "wrist": synced.frames["wrist"].data,
        "overhead": synced.frames["overhead"].data,
    },
    "depth": {
        "wrist": camera_wrist.read_depth().data,
    },
    "proximity": proximity_data,
    "state": robot_obs["state"],
})
```
The consumer always needs to know which sensor is which. A shared read() interface doesn't help here.

2. Lifecycle management. It saves repeating `__enter__/__exit__` for each sensor type. Convenient, but that's 4 lines of trivial boilerplate per sensor ABC. Not sure it's worth imposing a class hierarchy for that.

I think a lightweight `Device(Protocol)` for lifecycle management is enough. It gives us a shared contract (connect, disconnect, is_connected, context manager) for any code that needs to operate on devices generically , without imposing inheritance. Camera satisfies it structurally with no changes. Future sensor types will too, as long as they follow the same naming convention. We don't need to add it right now either; it's a non-breaking change that can be introduced when we add a new device type.

What do you think?

samet-akcay: > I think a lightweight Device(Protocol) for lifecycle management is enough. It gives us a shared contract (connect, disconnect, is_connected, context manager) for any code that needs to operate on devices generically , without imposing inheritance. Camera satisfies it structurally with no changes. Future sensor types will too, as long as they follow the same naming convention. We don't need to add it right now either; it's a non-breaking change that can be introduced when we add a new device type.

This would do the job.

My point was that we need to take that ask/request into account for the future. We don't have to do implement it now, but have some interface to follow. Protocol would be perfectly fine for this

samet-akcay (approved):

PR #3, opened 2026-03-16
feat(robot): SO101 implementation

maxxgx:
- Robot contract: protocol-based robot API.
- Implemented the SO101 driver base on `feetech-servo-sdk`
- Exposed SO101 through the robot package with lazy loading and added the so101 optional dependency.
- Added demo scripts for reading joints and moving joints on real hardware.
- Added unit tests covering the protocol, lifecycle utilities, conformance checks, and the SO101 driver with mocked hardware SDK interactions.


src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+    NUM_JOINTS: ClassVar[int] = 6
+    """Number of joints / servos on the SO-101."""
+
+    def __init__(
+        self,
+        port: str,
+        baudrate: int = 1_000_000,
+        role: str = "follower",
+        servo_ids: dict[str, int] | None = None,
+        calibration_path: str | Path | None = None,
```
MarkRedeman: In physical AI studio we store our calibration values in our SQLite database. We _also_ store them on disk but that's actually mainly because that's a requirement from lerobot's interface as it also requires a calibration path.

What would you think of changing this to instead accept a `SO101Calibratoin` (or some other name) that contains all calibration related info.

For convenience we could still expose a named static constructor:
```
robot =  SO101::fromCalibrationPath(..., calibration_path="...")
```

another possible benefit would be that in unit tests you wouldn't need to mock the filesystem.

maxxgx: Will add that, thanks!

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+            self._packet_handler,
+            _ADDR_GOAL_POSITION,
+            _LEN_GOAL_POSITION,
+        )
+
+        # Configure torque based on role ------------------------------------
+        self._set_torque(enabled=self.role == "follower")
+
+        logger.info(
+            "SO-101 connected on %s (role=%s, servos=%s)",
```
samet-akcay: not a loguru expert, but as far as I'm aware loguru favors `str.format` brace style, no?

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+    def _ticks_to_radians(self, ticks: np.ndarray) -> np.ndarray:
+        """Convert raw servo ticks to radians using calibration data.
+
+        Args:
+            ticks: Integer tick values, shape ``(6,)``.
+
+        Returns:
+            Float32 array of joint positions in radians, shape ``(6,)``.
+        """
+        assert self._calibration is not None  # noqa: S101
```
samet-akcay: would it be possible to avoid using assert? bandit does not like it, it vanishes under `python -0`
I would replace it with python exceptions?

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+            self._hold_position()
+
+        self._group_sync_read = None
+        self._group_sync_write = None
+        self._packet_handler = None
+        self._port_handler.closePort()
+        self._port_handler = None
+
+        logger.info("SO-101 disconnected from %s", self.port)
+
```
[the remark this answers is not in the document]
samet-akcay: I think this could be a valid issue indeed. 

We might need to wrap this with a try/except and do always cleanup

src/physicalai/robot/testing.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        robot: An object that is expected to satisfy the Robot protocol.
+        _num_steps: Number of observation/action round-trips to execute
+            (currently reserved for future use).
+    """
+    robot.connect()  # type: ignore[attr-defined]
+
+    obs = robot.get_observation()  # type: ignore[attr-defined]
+    assert isinstance(obs, dict), "get_observation() must return a dict"  # noqa: S101
+
```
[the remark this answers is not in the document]
samet-akcay: would it be dangerous to send zeros with`action = np.zeros(state_dim, dtype=np.float32)`, especially if a robot is not calibrated properly. 

how about echoing back to the current state?
Something like `action = obs["state"].copy()`?

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+        Raises:
+            ValueError: If ``role`` is not ``"leader"`` or ``"follower"``.
+        """
+        if role not in _VALID_ROLES:
+            msg = f"Invalid role {role!r}. Must be one of {sorted(_VALID_ROLES)}."
+            raise ValueError(msg)
+
+        self.port = port
+        self.baudrate = baudrate
+        self.role = role
```
samet-akcay: maybe make this property? we dont want it to be mutable right?

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+                "Leader arms are read-only for teleoperation."
+            )
+            raise RuntimeError(msg)
+
+        expected_shape = (self.NUM_JOINTS,)
+        if action.shape != expected_shape:
+            msg = f"Expected action shape {expected_shape}, got {action.shape}"
+            raise ValueError(msg)
+
+        ticks = self._radians_to_ticks(action) if self._calibration is not None else np.round(action).astype(np.int32)
```
samet-akcay: The uncalibrated path in `send_action` doesn't clamp tick values before they `hit _write_joint_positions`. The calibrated path does this via `np.clip` in `_radians_to_ticks`, but without calibration we just do `np.round(action).astype(np.int32)` and pass it straight through.


Should we clamp in the uncalibrated path too? Like,

`ticks = np.clip(np.round(action).astype(np.int32), 0, _TICKS_PER_REVOLUTION - 1)`

src/physicalai/robot/so101/so101.py
maxxgx: Yeah that true, fixed the formatting

src/physicalai/robot/so101/so101.py
maxxgx: Good point, added the clamping on both paths

src/physicalai/robot/so101/so101.py
maxxgx: Right, changed to property

src/physicalai/robot/so101/so101.py
maxxgx: Removed

src/physicalai/robot/so101/so101.py
maxxgx: Update: I made calibration required since I imagine we need that for inference. But I kept a uncalibrated path, mostly just for debug purposes

src/physicalai/robot/testing.py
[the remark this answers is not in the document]
maxxgx: Thanks, fixed

src/physicalai/robot/so101/so101.py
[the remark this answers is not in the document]
maxxgx: Done

src/physicalai/robot/so101/demo/README.md
```diff
@@ -0,0 +1,28 @@
...
+Read and display live joint positions. Torque is off — move the arm by hand to verify readings.
+
+```bash
+python -m physicalai.robot.so101.demo.read_joints --port /dev/ttyUSB0
+```
+
+## move_joints
+
+Move each joint ±50 ticks one at a time to verify actuation and wiring. Torque is released when done.
+
```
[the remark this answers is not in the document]
maxxgx: fixed

src/physicalai/robot/so101/demo/README.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+python -m physicalai.robot.so101.demo.move_joints --port /dev/ttyUSB0
+```
+
+Use `--offset 100` for larger movements or `--delay 1.0` for more time to observe.
```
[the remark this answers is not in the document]
maxxgx: fixed

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+            raise ValueError(msg)
+
+        self.port = port
+        self.baudrate = baudrate
+        self.role = role
+
+        # Servo ID mapping — default 1..6 in JOINT_ORDER
+        self.servo_ids: dict[str, int] = servo_ids or {
+            name: idx + 1 for idx, name in enumerate(self.JOINT_ORDER)
+        }
```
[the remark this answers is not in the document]
maxxgx: remove `servo_ids` (unnecessary)

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,470 @@
...
+            if not self._group_sync_read.addParam(servo_id):
+                msg = f"Failed to add servo {servo_id} to sync read group"
+                raise ConnectionError(msg)
+
+        self._group_sync_write = GroupSyncWrite(
+            self._port_handler,
+            self._packet_handler,
+            _ADDR_GOAL_POSITION,
+            _LEN_GOAL_POSITION,
+        )
```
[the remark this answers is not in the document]
maxxgx: addressed

src/physicalai/robot/so101/demo/__init__.py
samet-akcay, line 1: where do you think this folder would be the best located? 
here in the src, or would we have an `examples` directory under the root?

samet-akcay: It looks a bit odd to me to have the demos inside src

src/physicalai/robot/testing.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import annotations
+
+import time
+
+import numpy as np
+
+from physicalai.robot.utils import connect
+
+def check_robot_conformance(robot: object, _num_steps: int = 10) -> None:
```
samet-akcay: how about the following?

```diff
+ validate_robot # name of the function
- check_robot_conformance
```

and

```diff
+ validation.py # name of the file
- testing.py
```

pyproject.toml
```diff
@@ -108,10 +110,7 @@ max-args = 15
 convention = "google"
 
 [tool.ruff.lint.flake8-copyright]
-notice-rgx = """
-# Copyright \(C\) (\d{4}(-\d{4})?) Intel Corporation
-# SPDX-License-Identifier: Apache-2\.0
-"""
+notice-rgx = '# Copyright \(C\) (\d{4}(-\d{4})?) Intel Corporation\n# SPDX-License-Identifier: Apache-2\.0'
```
samet-akcay, line 114: is this definitely needed?

maxxgx: Looks like this is needed if we want to keep our copyright enforcement. I think it's worth to keep since it's easy to forget and omit the header

samet-akcay: hmm, that's peculiar. Because I introduced it in Anomalib and used in other repos as well. 

Not a big issue, just the previous version is the exact same format that improves the readability. Nothing wrong with the new one

src/physicalai/robot/testing.py
maxxgx: I've renamed it to `verify.py` and `verify_robot()` because the design already uses "validation" for runtime  checks against the policy manifest. Using `validation.py` here might create confusion, I think using verify keeps the two concepts distinct.

Current structure looks like:
```
src/physicalai/robot
├── __init__.py
├── connect.py
├── interface.py
├── so101
│   ├── __init__.py
│   └── so101.py
└── verify.py
```

src/physicalai/robot/so101/demo/__init__.py
maxxgx: Agree, I moved it to `examples/`

src/physicalai/robot/testing.py
samet-akcay: sure verify sounds good

src/physicalai/robot/so101/so101.py
```diff
@@ -0,0 +1,670 @@
...
+
+    NUM_JOINTS: ClassVar[int] = 6
+    """Number of joints / servos on the SO-101."""
+
+    def __init__(
+        self,
+        port: str,
+        calibration: SO101Calibration | str | Path | None,
+        baudrate: int = 1_000_000,
+        role: str = "follower",
```
samet-akcay: I think it would be more descriptive if we have one of the two:

```python
role: Literal["leader", "follower"]
```

or something like:

```python
role: str | Role = "follower"  # or whatever you named the StrEnum
...
role = Role(role)
```

maxxgx: Thanks, I've applied the suggestion using `Literal`

PR #4, opened 2026-03-23
feat: Add pre-commit checks

maxxgx:
Adds repository-wide pre-commit checks with `prek` and updates the codebase to pass them cleanly.

## Changes

- add `.pre-commit-config.yaml` and `.markdownlint.yaml`
- enable `ruff`, `pyrefly`, `uv lock`, `hadolint`, `markdownlint`, and `zizmor`
- update `pyproject.toml` and `uv.lock` for the new tooling
- refactor SO101 code by extracting calibration and constants modules
- clean up SO101 implementation, verification, tests, examples, and docs to satisfy linting, formatting, and type checks


.pre-commit-config.yaml
```diff
@@ -0,0 +1,56 @@
+# Copyright (C) 2025 Intel Corporation
```
samet-akcay:
```suggestion
# Copyright (C) 2026 Intel Corporation
```

pyproject.toml
```diff
@@ -97,7 +101,10 @@ extend-exclude = ["docs", "examples", "tests", "notebooks"]
 
 [tool.ruff.lint]
 select = ["ALL"]
-ignore = []
+ignore = [
+    "COM812",  # Conflicts with Ruff formatter
```
samet-akcay, line 105: is this a bad thing, why ignore?

pyproject.toml
```diff
@@ -97,7 +101,10 @@ extend-exclude = ["docs", "examples", "tests", "notebooks"]
...
+    "UP037",   # Removes quotes inside Literal[], breaking string literal values
```
samet-akcay: can we use noqa instead for this edge case, instead of completely ignoring this rule ?

pyproject.toml
maxxgx: COM812 is redundant/conflicts when Ruff formatter is enabled, and Ruff recommends disabling it in formatter-driven setups

pyproject.toml
maxxgx: Good point. I’ll replace the project-wide `UP037` ignore with a `noqa` on the affected lines

maxxgx: It seems `noqa` is not actually needed. Ruff removes the quotes only when you import `from __future__ import annotations`

samet-akcay: that's even better

samet-akcay (approved):

PR #7, opened 2026-03-27
docs: converged manifest format — dual component resolution

samet-akcay:
## Summary

Design documentation update for the converged manifest format used by both PhysicalAI and LeRobot inference exports. This PR documents the **dual component resolution** approach and removes the `_normalize_metadata()` migration shim in favor of a clean-cut nested Pydantic design.

**Requesting team feedback on the design before implementation begins.**

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| **Dual resolution** (`type` + `class_path`) | `type` + flat params for interoperability (LeRobot writes this). `class_path` + `init_args` for full-power PhysicalAI. Both resolve through the same `ComponentRegistry` → `instantiate_component()` pipeline. One if-check, not an if-chain per type. |
| **No `_normalize_metadata()`** | Since we're writing both the schema and loader simultaneously, there's no intermediate state requiring a flattening shim. Clean cut to nested Pydantic models. |
| **`from_legacy_metadata()` only** | Pre-manifest `metadata.yaml` files (early PhysicalAI exports) are the only backward compat path needed. |

## How to Review

1. Start with the **Executive Summary** and **Dual Component Resolution** table in `lerobot.md` (top of file)
2. Review **Section 3** (How PhysicalAI Loads) — this is the core design change
3. Check the **PhysicalAI-native manifest example** in `inferencekit.md` → Manifest Format section
4. Review **Section 10** (Migration) — significantly simplified

## Questions for Reviewers

- Does the dual resolution approach (type + class_path) make sense for your use cases?
- Any concerns about dropping `_normalize_metadata()` entirely?
- Is the migration path (Section 10) clear and sufficient?

## Related

- [Inference pipeline refactor PRs (merged)](https://github.com/samet-akcay/physical-ai-studio/pulls?q=is%3Apr+is%3Amerged+inference): Phases 1-4 complete
- Phases 5-6 implementation pending this design approval

### DIFF
--- a/docs/design/components/inferencekit.md
+++ b/docs/design/components/inferencekit.md
@@ -438,7 +438,7 @@ class Postprocessor(ABC):
 
 ### Manifest Format
 
-All exported models use a unified `manifest.json` format. The manifest uses a nested structure that mirrors the `InferenceModel` class hierarchy, with logical sections for policy identity, inference configuration, hardware, and metadata:
+All exported models use a unified `manifest.json` format. The manifest uses a nested structure that mirrors the `InferenceModel` class hierarchy, with logical sections for policy identity, model configuration, hardware, and metadata:
 
 ```text
 manifest.json
@@ -446,7 +446,7 @@ manifest.json
 ├── policy                  (identity — what policy is this?)
 │   ├── name                (human-readable name)
 │   └── source              (provenance: repo_id, class_path)
-├── inference               (InferenceModel — how to run it?)
+├── model                   (exported model — how to run it?)
 │   ├── n_obs_steps         (observation window)
 │   ├── runner              (execution pattern + params)
 │   ├── artifacts           (model files by named role)
@@ -468,7 +468,7 @@ manifest.json
       "class_path": "mypackage.policies.MyPolicy"
     }
   },
-  "inference": {
+  "model": {
     "n_obs_steps": 1,
     "runner": {
       "type": "action_chunking",
@@ -532,10 +532,10 @@ PhysicalAI can also write manifests using the full `class_path` + `init_args` fo
       "class_path": "physicalai.policies.act.policy.ACT"
     }
   },
-  "inference": {
+  "model": {
     "n_obs_steps": 1,
     "runner": {
-      "class_path": "action_chunking",
+      "class_path": "physicalai.inference.runners.ActionChunkingRunner",
       "init_args": {
         "chunk_size": 100,
         "n_action_steps": 100
@@ -554,7 +554,7 @@ PhysicalAI can also write manifests using the full `class_path` + `init_args` fo
       ],
       "preprocessors": [
         {
-          "class_path": "normalize",
+          "class_path": "physicalai.inference.preprocessors.StatsNormalizer",
           "init_args": {
             "mode": "mean_std",
             "stats_path": "stats.safetensors",
@@ -564,7 +564,7 @@ PhysicalAI can also write manifests using the full `class_path` + `init_args` fo
       ],
       "postprocessors": [
         {
-          "class_path": "denormalize",
+          "class_path": "physicalai.inference.postprocessors.StatsDenormalizer",
           "init_args": {
             "mode": "mean_std",
             "stats_path": "stats.safetensors",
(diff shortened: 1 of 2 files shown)


docs/design/components/inferencekit.md
```diff
@@ -438,77 +438,169 @@ class Postprocessor(ABC):
...
+      "model": "model.onnx"
+    },
+    "io": {
+      "inputs": [
+        {"name": "observation.image", "dtype": "float32", "shape": ["B", 3, 96, 96]},
+        {"name": "observation.state", "dtype": "float32", "shape": ["B", 14]}
+      ],
+      "outputs": [
+        {"name": "action", "dtype": "float32", "shape": ["B", 100, 14]}
+      ],
```
maxxgx: Do we really need `inputs` and `outputs` sections? This will conflict with robots and cameras, where we basically have the same shapes as here.

In case they are absolutely needed, I'd refactor hardware as:
```json
"hardware": {
  "robot_type": "SO-100",
  "cameras": ["top", "wrist"]
}
```

docs/design/components/inferencekit.md
```diff
@@ -438,77 +438,169 @@ class Postprocessor(ABC):
...
+      "preprocessors": [
+        {
+          "type": "normalize",
+          "mode": "mean_std",
+          "artifact": "stats.safetensors",
```
maxxgx:
```suggestion
          "stats_path": "stats.safetensors",
```

docs/design/components/inferencekit.md
```diff
@@ -438,77 +438,169 @@ class Postprocessor(ABC):
...
+          "features": ["observation.state"]
+        }
+      ],
+      "postprocessors": [
+        {
+          "type": "denormalize",
+          "mode": "mean_std",
+          "artifact": "stats.safetensors",
```
maxxgx:
```suggestion
          "stats_path": "stats.safetensors",
```

docs/design/components/inferencekit.md
maxxgx: Moreover, are the input shapes pre/post preproccessing/postprocessing?

docs/design/components/inferencekit.md
```diff
@@ -438,77 +438,169 @@ class Postprocessor(ABC):
...
-      }
-    }
-  ],
-  "postprocessors": [
-    {
-      "class_path": "mypackage.postprocessors.NMS",
+  "inference": {
+    "n_obs_steps": 1,
+    "runner": {
+      "class_path": "action_chunking",
```
maxxgx: I think for class_path + init_args, the path should always be the full python class path. Otherwise, it makes more sense to use type

docs/design/components/inferencekit.md
```diff
@@ -438,77 +438,169 @@ class Postprocessor(ABC):
...
+      "repo_id": "user/my_model",
+      "class_path": "mypackage.policies.MyPolicy"
+    }
   },
-  "artifacts": {
-    "onnx": "model.onnx"
+  "inference": {
+    "n_obs_steps": 1,
+    "runner": {
+      "type": "action_chunking",
```
maxxgx, line 454: If I understand correctly, the manifest supports two component formats:

- `type + flat params`: manifest exported from lerobot and can be read by physicalai.
- `class_path + init_args`: for custom components that aren't in the built-in registry. Only physicalai reads this.

This makes sense, type keeps the system portable, while class_path keeps the system open so 3rd parties can plug in custom runners/processors without needing a PR to physicalai

samet-akcay: yes this is the idea

docs/design/components/inferencekit.md
samet-akcay: This is actually a good suggestion, thanks

docs/design/components/inferencekit.md
```diff
@@ -438,195 +417,184 @@ class Postprocessor(ABC):
...
+          "mode": "mean_std",
+          "stats_path": "stats.safetensors",
+          "features": ["action"]
+        }
+      }
+    ]
+  },
+  "hardware": {
+    "robots": [],
+    "cameras": []
```
MarkRedeman: Can we include an example of what this looks like?
I'd like to use this as a reference for the naming mapping issue that Ronald spoke about during the weekly.

samet-akcay: Added concrete examples. Both manifest examples now show a filled-in hardware section. I was thinking about something like this, but we need to discuss together to settle on the design. @maxxgx 

```
 hardware: {
   robots: [
     {
       name: main,
       type: SO-100,
       state: {
         shape: [6], dtype: float32,
         order: [shoulder_pan, shoulder_lift, elbow_flex, wrist_flex, wrist_roll, gripper]
       },
       action: { ... same ... }
     }
   ],
   cameras: [
     {name: top, shape: [3, 480, 640], dtype: uint8},
     {name: wrist, shape: [3, 480, 640], dtype: uint8}
   ]
 }
```
 
The key thing for the naming mapping: name fields are logical names matching the keys used during training. At deployment, the user maps these to physical devices. The order field declares joint ordering explicitly — the runtime can compare it against the robot's actual joint order to catch mismatches at startup (critical for multi-arm setups).

maxxgx: Yeah I think this is the format we've settled on at the moment. Note it's possible that `state` might have a different shape compared to `action`, e.g. `state` might include additional sensor readings as well

PR #9, opened 2026-04-01
feat(capture): add UVC camera capture

maxxgx:
This PR includes implementation of `physicalai.capture`:
- step 1: core API, shared utilities, and hardware-free test scaffolding.
- step 2: add `UVCCamera` via omnicamera and v4l2 backends

Note: on linux we always use V4L2 discovery and camera controls because it provides richer information than omnicamera currently exposes.


src/physicalai/capture/cameras/uvc/_omnicamera.py
```diff
@@ -0,0 +1,269 @@
...
+                )
+            )
+        return controls
+
+    def apply_settings(self, settings: CameraSetting | list[CameraSetting]) -> None:
+        """Apply one or more camera settings.
+
+        Read-only, inactive, and valueless settings are silently skipped.
+        """
+        raise NotImplementedError
```
maxxgx, line 265: I'll add the implementation in a later PR. I need to check what nokhwa and omnicamera actually expose for camera controls

docs/design/components/camera-interface.md
```diff
@@ -627,33 +639,71 @@ line longer and gives the caller full control over error handling.
...
 ## Proposed Implementations
 
-### OpenCVCamera
+### UVCCamera
+
+USB webcams, built-in cameras, and UVC devices on **all platforms**.
+`UVCCamera` is a facade that delegates to a platform-specific backend:
+
+- **macOS/Windows/Linux**: `OmniCameraBackend` via [OmniCamera](https://github.com/ArendJanKramer/OmniCamera)
+- **Linux**(optional): `V4L2Camera` via native V4L2 ioctls
```
samet-akcay: If OmniCameraBackend and V4L2Camera same level backend entity, the way they are named looks inconsistent?

src/physicalai/capture/cameras/uvc/v4l2/_v4l2camera.py
samet-akcay, line 1: what is the difference between _camera.py and `_v4l2camera.py`?

src/physicalai/capture/cameras/uvc/_omnicamera.py
samet-akcay, line 1: `v4l2camera.py` lives under v4l2 as a backend, but `_omnicamera.py` lives in the same level in the uvc directory? would it be an idea to make this backend vs camera more distinct?

src/physicalai/capture/cameras/uvc/_camera.py
```diff
@@ -0,0 +1,196 @@
...
+
+    Args:
+        device: Unified device selector.
+            - On Linux (V4L2): ``0`` maps to ``/dev/video0``.
+            - On macOS/Windows (OmniCamera): ``0`` maps to OmniCamera index ``0``.
+        width: Requested frame width in pixels.
+        height: Requested frame height in pixels.
+        fps: Requested frames per second.
+        color_mode: Pixel format for returned frames.
+        backend: ``"auto"``, ``"v4l2"``, or ``"omnicamera"``.
```
samet-akcay: is `auto` intentional? We don't seem to handle auto?

src/physicalai/capture/cameras/uvc/_omnicamera.py
```diff
@@ -0,0 +1,269 @@
...
+
+if TYPE_CHECKING:
+    from physicalai.capture.discovery import DeviceInfo
+
+
+_MISSING_DEP_PKG = "omni_camera"
+_MISSING_DEP_EXTRA = "capture"
+
+try:
+    import omni_camera  # type: ignore[import-not-found]
```
samet-akcay: does this always avoid escaping the import issues, when omni_camera is not installed, but not used? like would we need lazy loading or importing inside the class methods ?

pyproject.toml
```diff
@@ -14,16 +14,21 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python :: 3.12",
 ]
-dependencies = ["numpy>=1.24", "loguru>=0.7"]
+dependencies = [
+    "numpy>=1.24",
+    "loguru>=0.7",
+    # target PyTurboJPEG 1.x for libjpeg-turbo 2.x compatibility
+    "PyTurboJPEG<2; sys_platform == 'linux'",
+    "omni_camera @ git+https://github.com/ArendJanKramer/OmniCamera.git@master",
```
samet-akcay, line 22: is this pypi friendly? if they haven't changed their policy, pypi blocks git installs due to security concerns?

pyproject.toml
```diff
@@ -14,16 +14,21 @@ classifiers = [
...
+    "pydantic==2.12.5",
```
samet-akcay: maybe we could flex this a little bit. Since this is gonna be a library, it should be as flexible as possible

samet-akcay: something like `"pydantic>=2.10,<3"`? or whichever base version you need

pyproject.toml
```diff
@@ -14,16 +14,21 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Programming Language :: Python :: 3.12",
 ]
-dependencies = ["numpy>=1.24", "loguru>=0.7"]
+dependencies = [
+    "numpy>=1.24",
+    "loguru>=0.7",
+    # target PyTurboJPEG 1.x for libjpeg-turbo 2.x compatibility
+    "PyTurboJPEG<2; sys_platform == 'linux'",
```
samet-akcay, line 21: do we need this as core dependency?

examples/capture/read_uvc_camera.py
samet-akcay, line 1: regarding the examples in general, would it be an idea to have notebooks instead of python scripts

maxxgx: That will nice and would make the example more impactful. Since this PR is already too big, I'll add them in a later PR

pyproject.toml
maxxgx: It's required by the V4L2Camera implementation as we need it to decode the frames. Of course, this is only used if user selects the v4l2 backend. I'm still not sure on the viability of this path, but I'd like to keep it for now. 

IMO using a native backend (v4l2 on linux) should not require to install optional dependencies

pyproject.toml
maxxgx: Good point. thanks. `"pydantic>=2.5"` looks good as it contains major performance improvements

pyproject.toml
maxxgx: It's a temporary solution. The plan to add Arend Jan's OmniCamera fork to pypi, probably under some other name since it's already used by the original repo.

src/physicalai/capture/cameras/uvc/_omnicamera.py
maxxgx: Updated. Actually I think we can import it normally

src/physicalai/capture/cameras/uvc/_camera.py
maxxgx: Nice catch, no `auto` allowed here

src/physicalai/capture/cameras/uvc/_omnicamera.py
maxxgx: Both `v4l2/_camera.py` (remove duplicate `v4l2/_v4l2camera.py`) and `_omnicamera.py` are peer implementations used by `UVCCameara`, which is a facade over them.

I put the V4L2 implementation in its own folder because it includes several V4L2-specific pieces, such as the ctypes mappings, to it's easier to manage as a small sub package. OmniCamera implementation is smaller, so I kept it in a single file for now

src/physicalai/capture/cameras/uvc/v4l2/_v4l2camera.py
maxxgx: A file rename went wrong 😅 

I've removed `_v4l2camera.py`

docs/design/components/camera-interface.md
maxxgx: Agree, I don't like the `Backend` suffix either. Initially I thought it could be confused with the omni_camera package. But I think it does matter since it's an meant to be an internal class.

I've renamed it to `OmniCamera`

ArendJanKramer (approved):

PR #107, opened 2026-05-08
chore(physicalai): copy source code from PAI studio

maxxgx:
- Copying code from PAI studio main branch.
- Includes some small style fixes (markdown, ruff, pyrefly)
- Fix bandit issue: see commit [address bandit B615](https://github.com/openvinotoolkit/physicalai/pull/107/commits/83990d14f433905b8b2901cb2ffecad810571846)

### DIFF
--- a/src/physicalai/inference/preprocessors/hf_tokenizer.py
+++ b/src/physicalai/inference/preprocessors/hf_tokenizer.py
@@ -1,0 +1,105 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Lambda preprocessor that wraps an arbitrary callable."""
+
+from __future__ import annotations
+
+import numpy as np
+
+from physicalai.inference.constants import TASK, TOKENIZED_PROMPT, TOKENIZED_PROMPT_MASK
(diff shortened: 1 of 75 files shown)


src/physicalai/inference/preprocessors/hf_tokenizer.py
```diff
@@ -29,17 +33,22 @@ def __init__(self, tokenizer_name: str, revision: str, max_token_len: int = 512)
...
         super().__init__()
+        if not _COMMIT_HASH_RE.fullmatch(revision):
+            msg = "revision must be an immutable Hugging Face commit hash (7-40 hex chars), not a branch or tag"
+            raise ValueError(msg)
+
         try:
             from transformers import AutoTokenizer  # noqa: PLC0415
 
-            self._tokenizer = AutoTokenizer.from_pretrained(
+            self._tokenizer = AutoTokenizer.from_pretrained(  # nosec B615
```
samet-akcay: so we need to ignore this line even we specify the hash instead of `main`?

maxxgx: Forgot to remove this, let me try without the suppression

Edit: ignore was not necessary

PR #113, opened 2026-05-14
chore(ci): enable secret scanning

AlexanderBarabanov:
This PR adds secret scanning into CI pipeline.


.semgrepignore
```diff
@@ -0,0 +1 @@
+**/uv.lock
```
samet-akcay: do we need formatting for these files ?

AlexanderBarabanov: thanks, updated

samet-akcay (approved):

PR #118, opened 2026-05-17
feat(runtime): basic PolicyRuntime implementation

### LINKED ISSUE #15: Epic: Runtime System MVP
## Summary
Implement the local runtime path described in the runtime-system design.

## Scope
Add `physicalai.runtime`, local sync execution, runtime queueing, config loading, and `physicalai run`.

## Status
Planned

## Dependencies
#16

## Related Issues
- #16 Inference Runtime Contract

maxxgx:
## Summary

This PR contains Phase 1 (of 2):
- Initial policy runtime and async execution (threaded).
- Thread safe action queue
- Chunk smoothing: replace or merge via lerp (linear interpolation)

Phase 2: https://github.com/openvinotoolkit/physicalai/pull/119

## Why

-

## Validation

-

## Breaking changes

- None

## Related issues

- Closes #


src/physicalai/inference/model.py
```diff
@@ -8,15 +8,19 @@
...
+from typing import TYPE_CHECKING, Any, Self, cast
 
 import yaml
 
 from physicalai.inference.adapters import adapter_registry, get_adapter
 from physicalai.inference.component_factory import instantiate_component, resolve_artifact
 from physicalai.inference.constants import ACTION
 from physicalai.inference.manifest import ComponentSpec, Manifest
-from physicalai.inference.runners import get_runner
+from physicalai.inference.runners import ActionChunking, get_runner
```
samet-akcay: Also note that @sovrasov created this PR on the studio side, which would conflict with this 
https://github.com/open-edge-platform/physical-ai-studio/pull/598

This is as part of the following `ActionChunkCursor` [design](https://github.com/open-edge-platform/physical-ai-studio/blob/main/physicalai/docs/design/components/runtime-system/policy_runtime_design.md#3-select_action-vs-predict_action_chunk)

examples/runtime/async_inference.py
```diff
@@ -0,0 +1,128 @@
...
+    print("Starting policy runtime...")
+    try:
+        stats = runtime.run(duration_s=args.duration_s)
+        print(f"\nDone — {stats.steps} steps, {stats.inference_count} inferences, {stats.total_holds} holds")
+    finally:
+        for name, cam in cameras.items():
+            cam.disconnect()
+            print(f"Camera '{name}' disconnected")
+        robot.disconnect()
+        print("Robot disconnected")
```
samet-akcay: `PolicyRuntime` already owns robot and cameras, so it might be an idea if it owns their connection lifecycle too?

Right now the example has to manually `connect()` / `disconnect()` everything around `runtime.run()`, which makes the API a bit tedious to setup.

Would it make sense to add `PolicyRuntime.connect()`, `disconnect()`, and `__enter__ `/ `__exit__` so callers can do:

```python
with runtime:
    stats = runtime.run(...)
```

That would make cleanup much safer and keep the example a lot cleaner?

 I’d still keep the user-facing camera discovery / debug printing in the example, but the actual resource ownership seems like it belongs in `PolicyRuntime`.

Thoughts?

maxxgx: Makes sense, I like the idea. Moved the robot/camera lifecycle inside the runtime

src/physicalai/inference/utils/action_cursor.py
maxxgx, line 1: I removed `ActionCursor`, as it's only a wrapper over `deque`. Because I think it can be confusing with the `ActionQueue`, as it's conceptually similar

src/physicalai/inference/preprocessors/pi05.py
```diff
@@ -135,7 +135,9 @@ def _preprocess_images(
             if img.ndim == max_image_dim:
                 img = img[:, -1, :, :, :]
 
-            if img.dtype != np.float32:
+            if img.dtype == np.uint8:
+                img = img.astype(np.float32) / 255.0
+            elif img.dtype != np.float32:
                 img = img.astype(np.float32)
```
maxxgx, line 141: I think, we need to normalized uint8, right @sovrasov?

examples/runtime/async_inference.py
```diff
@@ -0,0 +1,104 @@
...
+        cameras=cameras,
+        fps=args.fps,
+    )
+
+    try:
+        runtime.connect()
+    except Exception as e:
+        print(f"Failed to connect: {e}")
+        print("Available cameras:")
+        for driver, devices in discover_all().items():
```
samet-akcay, line 86: just a UX comment...

I was just reading this part without looking at the imports, and thought `discover_all` discovers everything including robots etc. Which name would be more descriptive? `discover_all` or `discover_cameras` or `discover_all_cameras` (verbose), or something else ?

samet-akcay: Would `with runtime` approach still work here? Or do you prefer to connect and disconnect explicitly for error management ?

src/physicalai/inference/utils/action_cursor.py
samet-akcay: `ActionCursor` was the internal helper to ensure the consistency. As long as we have consistent behaviour between `select_action` and `ActionQueue`, it is ok.

src/physicalai/robot/__init__.py
```diff
@@ -13,12 +13,20 @@
...
 from physicalai.robot.verify import verify_robot
 
+if TYPE_CHECKING:
+    from physicalai.robot.so101 import SO101 as SO101
+    from physicalai.robot.trossen import BimanualWidowXAI as BimanualWidowXAI
+    from physicalai.robot.trossen import WidowXAI as WidowXAI
+
 __all__ = [
     "Robot",
+    "RobotObservation",
```
samet-akcay, line 29: if we introduce `Observation` object into this repo, we might need to revisit this part. But it is currently out of scope

maxxgx: Agreed

src/physicalai/inference/preprocessors/pi05.py
sovrasov: Right, the expected fp range is [0;1]

examples/runtime/async_inference.py
maxxgx: Good point, I think `discover_cameras` seems more fitting. Will keep it in mind for future improvements.

Should still work with the context manager. I'm updating the example to make it more generic in the next PR

src/physicalai/runtime/__init__.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2025-2026 Intel Corporation
```
maxxgx: Updated

src/physicalai/inference/model.py
```diff
@@ -207,31 +206,33 @@ def select_action(self, observation: dict[str, np.ndarray]) -> np.ndarray:
...
         Args:
             observation: Observation dict mapping names to numpy arrays.
 
         Returns:
-            Chunk of actions to execute.
+            2-D action chunk with shape ``(chunk_size, action_dim)``.
         """
         outputs = self(observation)
-        return outputs[ACTION]
+        actions = np.asarray(outputs[ACTION])
```
sovrasov: Are there any problems with actions being something other than `np.ndarray`? Strict numpy compatibility should be provided by runtime adapters we have

src/physicalai/inference/model.py
```diff
@@ -207,31 +206,33 @@ def select_action(self, observation: dict[str, np.ndarray]) -> np.ndarray:
...
+        # Drop the adapter's batch dimension so the return value matches
+        # the documented contract: a 2-D (chunk, action_dim) sequence.
+        return np.atleast_2d(np.squeeze(actions))
```
sovrasov: Now non-trivial batch dimension is now prohibited, and it looks like it'd be good to have a clear error message here instead of relying on np. Technically, it's still possible to do batched inference with OV, but most likely it can happen in torch under the good, especially in different evaluation scenarios

samet-akcay (dismissed):
would be good to add your inference demo to our main readme file. Does not have to be in this pr, but we should have it :)

src/physicalai/inference/model.py
maxxgx: Thanks, added error message

src/physicalai/inference/model.py
maxxgx: Right, removed redundant `np.asarray`

sovrasov (approved):

PR #123, opened 2026-05-19
chore(config): move shared config primitives into runtime package

samet-akcay:
## Summary

- add a runtime-owned `physicalai.config` package for shared configuration utilities
- move `ComponentSpec` out of `physicalai.inference.manifest` while keeping the old import path as a compatibility re-export
- update runtime inference code to use the shared config primitive directly

## Why

`Config`, ComponentSpec` and the generic config helpers need to live in the torch-free runtime package so both the runtime repo and the training library can use the same primitive. This keeps configuration ownership in one place and removes the long-term duplicate `physicalai.config` implementation.

## Compatibility

This PR keeps the existing import path working:

```python
from physicalai.inference.manifest import ComponentSpec
```

New code can use:

```python
from physicalai.config import ComponentSpec
```

## ⚠️ Follow-up

This is the first half of the migration. The paired training-library PR removes the duplicate `library/src/physicalai/config` package and switches library imports to the runtime-owned `physicalai.config` module.

### DIFF
--- a/src/physicalai/config/__init__.py
+++ b/src/physicalai/config/__init__.py
@@ -1,0 +1,11 @@
+# Copyright (C) 2025-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Configuration primitives shared by runtime and training packages."""
+
+from physicalai.config.base import Config
+from physicalai.config.component import ComponentSpec
+from physicalai.config.instantiate import instantiate_obj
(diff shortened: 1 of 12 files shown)


samet-akcay:
Just reading through https://jsonargparse.readthedocs.io/en/stable/ and feel like we could re-write all of this stuff using just jsonargparse

PR #128, opened 2026-05-21
feat(runtime): implement Real-Time Chunking (RTC) execution strategy

eugene123tw:
## Summary

- Implement Real-Time Chunking (RTC) execution strategy for continuous, non-blocking robot control with async background inference.

## Why

- Standard synchronous inference blocks the control loop, causing jitter at high FPS. RTC decouples inference from action dispatch by running predictions in a daemon thread and serving actions from a dual-track queue — enabling smooth 30 Hz+ robot control even when individual inferences take hundreds of milliseconds.

## Validation

- Manual testing with Pi0.5 RTC model on SO-101 robot at 30 FPS
- Warmup inference correctly resets the latency tracker to discard OpenVINO first-run compilation overhead
- Thread safety verified via `threading.Lock` on all shared state (observation slot, action queue)
- Error recovery: background thread tolerates up to 10 consecutive inference failures before halting

## Breaking changes

- None

## Related issues

- None

### Changes Made

| File | Description |
|------|-------------|
| rtc_execution.py | New `RTCExecution` class — background daemon thread continuously predicts action chunks, injects RTC inputs (noise, `prev_chunk_left_over`, `inference_delay`, `max_guidance_weight`, `execution_horizon`), and merges results into the dual-track queue |
| _rtc_action_queue.py | New `RTCActionQueue` — thread-safe dual-track queue storing both raw (for feedback) and postprocessed (for robot) actions, with cursor-based consumption and latency-aware trim on merge |
| rtc_latency.py | New `RTCLatencyTracker` callback — sliding-window latency measurement with `compute_delay(fps)` for the RTC model's `inference_delay` input |
| __init__.py | Export `RTCExecution` and `RTCActionQueue` |
| __init__.py | Export `RTCLatencyTracker` |
| runtime.py | Accept `RTCActionQueue` in `PolicyRuntime` type signature |
| _shared_camera.py | Minor fix: explicitly set `sample = None` after failed reads |
| rtc_inference.py | End-to-end example demonstrating RTC inference with Pi0.5 on SO-101 |

### Architecture

```mermaid
graph LR
    A[Main Thread - 30Hz tick] -->|publish obs| B[Observation Slot]
    B -->|read obs| C[RTC Daemon Thread]
    C -->|model inference| D[InferenceModel]
    D -->|raw actions| C
    C -->|merge| E[RTCActionQueue]
    A -->|pop 1 action| E
    E -->|action| F[Robot]
    C -->|latency| G[RTCLatencyTracker]
    G -->|compute_delay| C
```

### DIFF
--- a/src/physicalai/capture/transport/_shared_camera.py
+++ b/src/physicalai/capture/transport/_shared_camera.py
@@ -270,6 +270,7 @@ class SharedCamera(Camera):
             if sample is None:
                 break
             newest_sample = sample
+            sample = None
 
         if newest_sample is not None:
             header, frame = self._decode_sample(newest_sample)
@@ -297,6 +298,7 @@ class SharedCamera(Camera):
             if sample is None:
                 break
             newest_sample = sample
+            sample = None
 
         if newest_sample is not None:
             header, frame = self._decode_sample(newest_sample)
--- a/src/physicalai/capture/transport/_shared_camera.py
+++ b/src/physicalai/capture/transport/_shared_camera.py
@@ -270,6 +270,9 @@ class SharedCamera(Camera):
             if sample is None:
                 break
             newest_sample = sample
+            # Explicitly clear 'sample' to release reference. Otherwise, Python keeps
+            # it alive during the next receive() call, triggering iceoryx2 subscriber warnings/errors,
+            # especially when run in an IDE where debuggers hold onto local frames.
             sample = None
 
         if newest_sample is not None:
@@ -277,6 +280,8 @@ class SharedCamera(Camera):
             self._last_header = header
             self._latest = frame
             self._check_config_match(header)
+            # Explicitly clear 'newest_sample' to release the iceoryx2 sample borrow under IDE/debugger.
+            newest_sample = None
 
         if self._latest is None:
             msg = "no frame available"
@@ -298,6 +303,9 @@ class SharedCamera(Camera):
             if sample is None:
                 break
             newest_sample = sample
+            # Explicitly clear 'sample' to release reference. Otherwise, Python keeps
+            # it alive during the next receive() call, triggering iceoryx2 subscriber warnings/errors,
+            # especially when run in an IDE where debuggers hold onto local frames.
             sample = None
 
         if newest_sample is not None:
@@ -305,6 +313,8 @@ class SharedCamera(Camera):
             self._last_header = header
             self._latest = frame
             self._check_config_match(header)
+            # Explicitly clear 'newest_sample' to release the iceoryx2 sample borrow under IDE/debugger.
+            newest_sample = None
 
         if self._latest is None:
             msg = "no frame available"
--- a/src/physicalai/runtime/rtc_execution.py
+++ b/src/physicalai/runtime/rtc_execution.py
@@ -155,10 +155,10 @@ class RTCExecution(Execution):
             rtc_config = model.manifest.model_extra.get("rtc", {}) if hasattr(model, "manifest") else {}
             if isinstance(rtc_config, dict) and "chunk_size" in rtc_config:
                 self._chunk_size = int(rtc_config["chunk_size"])
-            elif hasattr(model, "chunk_size"):
+            elif model.chunk_size > 1:
                 self._chunk_size = model.chunk_size
             else:
-                self._chunk_size = 50
+                self._chunk_size = 50  # fallback default to Pi05 chunk size
 
         # 2. Infer max_action_dim
         if self._max_action_dim_param is not None:
--- a/src/physicalai/runtime/rtc_execution.py
+++ b/src/physicalai/runtime/rtc_execution.py
@@ -60,6 +60,9 @@ class RTCExecution(Execution):
         queue_threshold: Re-infer when queue drops below this level.
         latency_tracker: Callback that measures inference latency.
             If None, delay defaults to 0.
+        warmup_inferences: Number of initial inferences treated as warmup.
+            The latency tracker is reset after these to discard
+            compilation/kernel-build overhead (e.g. OpenVINO first-run).
         postprocessors: Denormalization pipeline applied to raw actions.
             These run in the background thread to produce the processed
             track stored in the queue.
@@ -74,6 +77,7 @@ class RTCExecution(Execution):
         max_guidance_weight: float = 10.0,
         queue_threshold: int | None = None,
         latency_tracker: RTCLatencyTracker | None = None,
+        warmup_inferences: int = 2,
         postprocessors: list[Postprocessor] | None = None,
     ) -> None:
         self._chunk_size = chunk_size
@@ -83,6 +87,7 @@ class RTCExecution(Execution):
         self._max_guidance_weight = max_guidance_weight
         self._queue_threshold = queue_threshold if queue_threshold is not None else execution_horizon * 3
         self._latency_tracker = latency_tracker
+        self._warmup_inferences = max(1, warmup_inferences)
         self._postprocessors: list[Postprocessor] = postprocessors or []
 
         self._rtc_queue: RTCActionQueue | None = None
@@ -154,7 +159,7 @@ class RTCExecution(Execution):
             self._obs_slot = deepcopy(sample_observation)
 
         # Wait for the first chunk with a generous timeout
-        if not self._first_chunk_ready.wait(timeout=30.0):
+        if not self._first_chunk_ready.wait(timeout=120.0):
             if self._death_cause is not None:
                 msg = f"RTC thread died during warmup: {self._death_cause}"
                 raise WorkerDiedError(msg) from self._death_cause
@@ -239,6 +244,15 @@ class RTCExecution(Execution):
 
             self._inference_count += 1
 
+            # Reset latency tracker after warmup inferences to discard
+            # compilation overhead (e.g. OpenVINO first-run latency).
+            if self._inference_count <= self._warmup_inferences and self._latency_tracker is not None:
+                self._latency_tracker.on_reset()
+                logger.info(
+                    "Warmup inference %d/%d complete (%.2fs) — latency tracker reset",
+                    self._inference_count, self._warmup_inferences, elapsed,
+                )
+
             # Extract raw actions: (1, chunk_size, action_dim) → (chunk_size, action_dim)
             raw_actions = outputs["action"]
             if raw_actions.ndim == 3:  # noqa: PLR2004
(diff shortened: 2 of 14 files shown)


src/physicalai/capture/transport/_shared_camera.py
```diff
@@ -270,6 +270,7 @@ def read(self, timeout: float = 2.0) -> Frame:
             if sample is None:
                 break
             newest_sample = sample
+            sample = None
```
maxxgx, line 276: I think both `newest_sample` and `sample` are local variables and do not hold the shared memory borrow after the function returns. So `sample=None` should not be necessary.

`self._held_sample` is where the reference is kept until the next frame

src/physicalai/runtime/runtime.py
```diff
@@ -74,7 +75,7 @@ def __init__(  # noqa: D107
         execution: Execution,
         fps: float,
         cameras: Mapping[str, Camera] | None = None,
-        action_queue: ActionQueue | None = None,
+        action_queue: ActionQueue | RTCActionQueue | None = None,
```
maxxgx: We could have an `ActionQueue` protocol, then we won't need to change the runtime API for new queue implementation which could be useful for users as well if they want to use their own class.

```
class ActionQueue(Protocol):
    def pop(self) -> np.ndarray | None: ...
    @property
    def remaining(self) -> int: ...
    @property
    def consecutive_holds(self) -> int: ...
    @property
    def total_holds(self) -> int: ...
    @property
    def total_pops(self) -> int: ...
    def below_threshold(self, threshold: int) -> bool: ...
    def clear(self) -> None: ...
```

This would require the current `ActionQueue` class to be renamed, e.g. `ChunkedActionQueue`

src/physicalai/runtime/rtc_execution.py
```diff
@@ -0,0 +1,334 @@
...
+        self,
+        chunk_size: int = 50,
+        execution_horizon: int = 10,
+        fps: float = 30.0,
+        max_action_dim: int = 32,
+        max_guidance_weight: float = 10.0,
+        queue_threshold: int | None = None,
+        latency_tracker: RTCLatencyTracker | None = None,
+        warmup_inferences: int = 2,
+        postprocessors: list[Postprocessor] | None = None,
```
maxxgx, line 88: This is a lot of parameters, is it possible to infer any of these from the manifest, like you already did for `postprocessors` in the example script?

Perhaps, we could a RTC section in the manifest with `chunk_size`, `max_action_dim`? @samet-akcay 

I guess `execution_horizon, max_guidance_weight, queue_threshold` can stay and be tuned by the uesr. Though, some indication of range and explanation of a specific value does would be great to add.

src/physicalai/runtime/rtc_execution.py
```diff
@@ -0,0 +1,334 @@
...
+    ) -> None:
+        self._chunk_size = chunk_size
+        self._execution_horizon = execution_horizon
+        self._fps = fps
+        self._max_action_dim = max_action_dim
+        self._max_guidance_weight = max_guidance_weight
+        self._queue_threshold = queue_threshold if queue_threshold is not None else execution_horizon * 3
```
maxxgx: `queue_threshold` and `execution_horizon` seem to be tightly coupled. Instead of having both arguments, could we only keep one (e.g. `execution_horizon`)?

Essentially queue_threshold is related to `execution_horizon`  + ` inference latency`?

maxxgx:
Some observations after testing on SO101:
- first few actions tend to move the robot very fast, almost violently. Maybe related to the warmup phase?
- with RTC, the robot is way less jittery and more precise
- in some scenarios, the model gets stuck by going back and forth without progress. Manually handing the object to the gripper helps to reset it. However, I'm not sure if it's RTC or model related.

src/physicalai/capture/transport/_shared_camera.py
eugene123tw: Local variables are normally cleaned up when a function returns, there are two reasons why explicitly setting `sample = None` and `newest_sample = None` is necessary here:

1. **reference overlap during loop iterations**:
   Inside the `while True:` loop, if we do not explicitly clear `sample` before the next iteration, the previous iteration's `sample` object is **still alive** when `self._subscriber.receive()` is called. 
   
   Because python evaluates the right-hand side (`self._subscriber.receive()`) *before* binding the result to the left-hand variable (`sample`), iceoryx2's C-extension sees two borrowed samples active at the exact same moment on python stack. Since iceoryx2's subscriber queues have extremely small default buffer capacities (typically 1 to 2 slots), this overlap triggers warnings (like `FailedToDeliverSignal`) or starves the buffer. Explicitly setting `sample = None` before the next `.receive()` ensures that only one raw sample remains borrowed at any point in the loop.

2. **IDE and Debugger frame retention**:
   When I ran the code inside IDEs or attaching an interactive debugger, the IDE's frame-evaluation and watch engines cache local stack frames and local variables (`sample`, `newest_sample`). 
   
   These debugger references keep the underlying Python objects alive even after the function returns, preventing the garbage collector from automatically releasing the iceoryx2 shared-memory borrows. Setting them to `None` breaks the binding instantly, guaranteeing that the shared-memory slots are returned to the pool immediately.

src/physicalai/runtime/rtc_execution.py
eugene123tw: I have refactored configuration handling and updated the model-loading logic. Here is how I addressed your feedback:

1. **auto-discovery of parameters**:
   * **`chunk_size` and `max_action_dim`**: Both are now optional in `RTCExecution`.
     * They are automatically read from the model manifest's `rtc` section (e.g. `model_extra.get("rtc", {})`). 
     * As a fallback, `chunk_size` is derived from model metadata, and `max_action_dim` is inferred from the robot's action shape defined in the manifest.
   * **`postprocessors`**: They are now auto-discovered from the `InferenceModel` during `.start()`. This migrates them into the background thread automatically without requiring manually loading them in the example script.

2. **dynamic queue_threshold calculation**:
   * I made `queue_threshold` optional. If not manually specified, it is dynamically computed. This automatically adjusts on-the-fly based on the real-time measured latency of the background thread and the loop's `fps`.

3. **detailed docstrings & code ranges**:
   * I updated `RTCExecution`'s docstring with explicit ranges and explanations for each tunable runtime parameter:
     * **`execution_horizon`** *(range: typically 5 to 15, default 10)*: Determines fresh actions executed per chunk. Lower values improve responsiveness/throughput but consume more CPU; higher values provide a buffer against execution/inference latency spikes.
     * **`max_guidance_weight`** *(range: typically 1.0 to 15.0, default 10.0)*: Controls the Classifier-Free Guidance (CFG) scale weight injected into diffusion/denoising models.
     * **`queue_threshold`**: Re-inference trigger point.

4. **simplified example script**:
   * I updated rtc_inference.py to showcase the simplified interface. Now, loading the model and starting execution is down to just a few lines with none of these parameters needing to be passed manually unless you want to override them."

eugene123tw:
> * first few actions tend to move the robot very fast, almost violently. Maybe related to the warmup phase?

This is because During the first inference step (warmup), the robot doesn't have a previous movement trajectory. Because of this, the action queue has no historical actions to return and `get_left_over()` evaluates to None. I set the guidance and execution_horizon to 0 when prev_chunk is None

> in some scenarios, the model gets stuck by going back and forth without progress. Manually handing the object to the gripper helps to reset it. However, I'm not sure if it's RTC or model related.

There's a few potential way to potentially alleviate it 
1. Relative action (not implemented yet)
2. I added a `LowPassFilterCallback` to run it: `python examples/runtime/rtc_inference.py --model ./model_path --low-pass-alpha 0.5 ...`

src/physicalai/capture/transport/_shared_camera.py
maxxgx: > Inside the while True: loop, if we do not explicitly clear sample before the next iteration, the previous iteration's sample object is still alive when self._subscriber.receive() is called.

> Because python evaluates the right-hand side (self._subscriber.receive()) before binding the result to the left-hand variable (sample), iceoryx2's C-extension sees two borrowed samples active at the exact same moment on python stack

I think this is true in either case, you still have `newest_sample` holding the same object. When `self._subscriber.receive()` is called on the next iteration, iceoryx2 sees 1 active borrow regardless of whether sample was set to None:
- with `sample = None`: refcount drops from 2→1, but `newest_sample` keeps the borrow alive
- without `sample = None`: refcount stays at 2, but it's the same borrow (same SHM slot)

However, good point on the debugger issue. I think Arend Jan hit something similar when running in PyCharm debugger, so having `sample = None` makes sense. Thanks for clarifying

src/physicalai/runtime/rtc_execution.py
```diff
@@ -0,0 +1,402 @@
...
+
+        # 1. Infer chunk_size
+        if self._chunk_size_param is not None:
+            self._chunk_size = self._chunk_size_param
+        else:
+            rtc_config = model.manifest.model_extra.get("rtc", {}) if hasattr(model, "manifest") else {}
+            if isinstance(rtc_config, dict) and "chunk_size" in rtc_config:
+                self._chunk_size = int(rtc_config["chunk_size"])
+            elif hasattr(model, "chunk_size"):
+                self._chunk_size = model.chunk_size
```
maxxgx: It seems that `model.chunk_size` is 1 with my exported OV model, there's no  chunk size yet in the manifest so its using the default property value `1`. 

```suggestion
            elif model.chunk_size > 1:
                self._chunk_size = model.chunk_size
```
It's not elegant but we first need to add `chunk_size` in the manifest

src/physicalai/runtime/rtc_execution.py
```diff
@@ -0,0 +1,402 @@
...
+    Args:
+        chunk_size: Number of actions per model output chunk. If None, is
+            automatically inferred from the model's manifest or model metadata.
+        execution_horizon: Fresh actions to execute per chunk (range: typically
+            5 to 15, default 10). Low values speed up response but consume more
+            CPU, high values tolerate higher latency spikes.
+        fps: Robot control frequency in Hz.
+        max_action_dim: Model's internal action dimension (for noise/padding).
+            If None, is automatically inferred from the model's manifest or
+            defaulted to 32.
```
maxxgx, line 61: Why not always infer from the model? In what scenario would user need to set it?

edit: We can improve this later

maxxgx (approved):

src/physicalai/runtime/rtc_execution.py
eugene123tw: At the moment, InferenceModel and Execution are independent, and have no dependency on each other. So it's not possible to get that info directly from OV model input to Execution. We also don't require saving extra RTC information when exporting model, so manifest stays clean. I don't think it's a huge issue as it could stay `None` and work fine without explicitly specifying unless needed.

MarkRedeman (approved):

PR #134, opened 2026-05-26
feat: inference performance benchmark

sovrasov:
## Summary

- introduce `physicalai.benchmark.perf` module that contains inference latency benchmark
- adds several entities describing input specifications for the `InferenceModel`
- `InferenceModel` reads full input features information from the manifest 
- update docs

## Why

- A quick inference benchmark allows to check inference latency on target hardware without setting up any complicated pipelines involving real robots and cameras

## Breaking changes

- None

## Example
```python
import argparse
import json

from physicalai.benchmark.perf import InferenceBenchmark
from physicalai.inference import InferenceModel
import numpy as np


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("export_dir", help="Path to the exported policy directory.")
    parser.add_argument("--max-iters", type=int, default=1000)
    parser.add_argument("--warmup-iters", type=int, default=1)
    parser.add_argument("--max-duration", type=int, default=60000, help="Wall-clock budget (ms).")
    args = parser.parse_args()

    model = InferenceModel(args.export_dir, device="CPU")
    benchmark = InferenceBenchmark(
        max_iters=args.max_iters,
        warmup_iters=args.warmup_iters,
        max_duration=args.max_duration,
    )
    metrics = benchmark.run(model)
    print(json.dumps(metrics, indent=2))


if __name__ == "__main__":
    main()
```


docs/explanation/inference.md
```diff
@@ -53,3 +53,27 @@ PolicyRuntime
   -> ActionQueue.pop_or_none()
   -> robot.send_action(action)
 ```
+
+## Performance Evaluation
+
+`InferenceBenchmark` measures per-chunk latency of an `InferenceModel` outside the runtime loop. It is intended for backend, device, and export-configuration comparisons; it is not a robot-loop profiler.
```
samet-akcay: Does this class do anything else than latency? If not, would it be an idea to name it as `LatencyBenchmark`? or `InferenceLatencyBenchmark`?

src/physicalai/benchmark/perf/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Performance benchmarks for physicalai components.
+
+Public API::
+
+    from physicalai.benchmark.perf import InferenceBenchmark
+"""
+
+from physicalai.benchmark.perf.inference_benchmark import InferenceBenchmark
```
samet-akcay: nitpick.. packages/sub-packages in physicalai follows the full name but not abbreviation. to follow the pattern, would it be an idea to name `perf` as `performance`

docs/explanation/inference.md
maxxgx: Does adding other metrics like memory/VRAM usage and power consumption make sense?

No need to be add them in this PR

sovrasov: That makes sense, let's see if we'll be able to find a reliable source for this data. For instance, OV don't add this into their benchmark despite heaving tighter relationships with driver/system utilities teams. For now, I'll rename

src/physicalai/benchmark/perf/__init__.py
sovrasov: Done

docs/explanation/inference.md
sovrasov: > InferenceLatencyBenchmark

Renamed to `InferenceLatencyBenchmark`

maxxgx (approved):

docs/explanation/inference.md
```diff
@@ -53,3 +53,27 @@ PolicyRuntime
   -> ActionQueue.pop_or_none()
   -> robot.send_action(action)
 ```
+
+## Performance Evaluation
+
+`InferenceLatencyBenchmark` measures per-chunk latency of an `InferenceModel` outside the runtime loop. It is intended for backend, device, and export-configuration comparisons; it is not a robot-loop profiler.
+
+```python
+from physicalai.benchmark.performance.inference_benchmark import InferenceLatencyBenchmark
```
samet-akcay, line 62: how about `inference_latency` instead of `inference_benchmark` as we are already in `benchmark` module so `from physicalai.benchmark.performance.inference_benchmark` looks like a repetition?

samet-akcay: do we plan on adding any other benchmark into `performance`?

PR #149, opened 2026-06-01
feat: add a universal resize preprocessing node

sovrasov:
## Summary

- resize node supports two extra parameters: padding and keep aspect ratio
- can be used for instance for ACT 

## Why

- flexible parametrized preprocessing node allows more flexibility for enabling new policies without modifying inference

## Validation

- Check against the original ACT torch resize implementation

## Breaking changes

- None

### DIFF
--- a/src/physicalai/inference/preprocessors/resize.py
+++ b/src/physicalai/inference/preprocessors/resize.py
@@ -101,8 +101,8 @@ class ResizePreprocessor(Preprocessor):
 
         if self._keep_aspect_ratio:
             ratio = max(cur_width / target_width, cur_height / target_height)
-            resized_height = min(int(cur_height / ratio), target_height)
-            resized_width = min(int(cur_width / ratio), target_width)
+            resized_height = max(1, min(int(cur_height / ratio), target_height))
+            resized_width = max(1, min(int(cur_width / ratio), target_width))
         else:
             resized_height = target_height
             resized_width = target_width
@@ -119,6 +119,9 @@ class ResizePreprocessor(Preprocessor):
         pad_bottom = pad_height - pad_top
         pad_left = pad_width // 2
         pad_right = pad_width - pad_left
+
+        if pad_height == 0 and pad_width == 0:
+            return img
 
         return np.pad(
             img,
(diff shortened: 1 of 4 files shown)


src/physicalai/inference/preprocessors/resize.py
```diff
@@ -0,0 +1,146 @@
...
+        pad_height = target_height - resized_height
+        pad_width = target_width - resized_width
+        pad_top = pad_height // 2
+        pad_bottom = pad_height - pad_top
+        pad_left = pad_width // 2
+        pad_right = pad_width - pad_left
+
+        return np.pad(
+            img,
+            ((0, 0), (0, 0), (pad_top, pad_bottom), (pad_left, pad_right)),
```
maxxgx, line 135: Is padding always `0`? If not, we could add a `padding_mode` parameter

src/physicalai/inference/preprocessors/resize.py
```diff
@@ -0,0 +1,146 @@
...
+        padding: Whether to pad images to reach the target resolution.
+        keep_aspect_ratio: Whether to preserve the aspect ratio when resizing.
+    """
+
+    def __init__(
+        self,
+        image_resolution: tuple[int, int],
+        *,
+        padding: bool = True,
+        keep_aspect_ratio: bool = True,
```
maxxgx: Maybe we can consider a cleaner way by defining:
```python
class ResizeMode(StrEnum):
    STRETCH   = "stretch"    # distort to exact target, no padding
    FIT       = "fit"        # preserve AR, output ≤ target (no padding)
    LETTERBOX = "letterbox"  # preserve AR + pad to exact target
```

Replacing the combo `padding` + `keep_aspect_ratio`.

maxxgx (dismissed):

src/physicalai/inference/preprocessors/resize.py
sovrasov: Added `pad_value` parameter to customize padding, but it still centers the output

maxxgx (dismissed):

src/physicalai/inference/preprocessors/resize.py
sovrasov: In our scenario, "fit" doesn't make sense, but others do. I added a enum: it'd be more convenient to add model unique modes, like "SMOLVLA" or similar

maxxgx (approved):

PR #150, opened 2026-06-02
fix(dep): support for python 3.11 and error for missing intel XPU dependencies

maxxgx:
## Summary

- Lower Python floor to **3.11** (was 3.12); classifiers list 3.11/3.12/3.13/3.14.
- Migrate `from typing import override` → `from typing_extensions import override` (13 files); add `typing-extensions>=4.5` dep.
- Add clear error and installation instruction for missing intel GPU plugin.
- Trossen on Python 3.14: `[trossen]` extra fails loudly; `[robots]` aggregator skips it via env marker; `WidowXAI` raises `ImportError` if `trossen-arm` is missing.

## Why

- Python 3.11 allows broader install reach (Debian 12, RHEL 9, ROS 2 Humble, JetPack 6).
- OpenVINO fails with opaque `RuntimeError` on GPU when plugin is missing.
- `trossen-arm` has no 3.14 wheel — project-wide `requires-python` cap would block unrelated users; silent skip hid the failure until runtime.

## Validation

- `uv run prek --all-files` → all hooks pass.
- Pyrefly / ruff delta vs. baseline: **0 new errors**.
- Unit tests on Python 3.11: **752 passed, 1 skipped**.
- Unit tests on Python 3.14 (`[capture] [robots] [observer-rerun]`): **723 passed, 3 skipped** (expected `trossen-arm`).
- Manually verified the new OpenVINO GPU error path on a host without the Intel compute runtime.

## Breaking changes

- None

## Related issues

- Closes #


docs/getting-started/installation.md
```diff
@@ -31,6 +31,29 @@ Training commands may be provided by a separate training distribution or by plug
 When installed together with the coordinated `physicalai-train` release, the
 shared `physicalai` executable also exposes training and benchmark subcommands.
 
+## OpenVINO GPU (optional)
```
sovrasov: Intel GPU runtime? The driver installation is common for every backend that relies on iGPU

pyproject.toml
```diff
@@ -27,9 +30,10 @@ dependencies = [
     "num2words>=0.5.14,<0.6.0",
     "safetensors>=0.4.3,<1.0.0",
     "onnxruntime",
-    "openvino==2026.1",
+    "openvino==2026.0",
```
sovrasov: Why do we need to pin 2026.0? Can we allow 2026.1 as well?

maxxgx: Yes, I was experimenting and forgot to revert

docs/getting-started/installation.md
maxxgx: Updated

sovrasov (approved):

PR #158, opened 2026-06-09
POC: add runtime teleoperation policy

MarkRedeman:
Reopening #131 on main branch.

TL/DR: implement a model policy whose predicted actions are the observed state from a leader robot.

```
class TeleoperatorPolicy:
    def predict_action_chunk(self, observation: dict[str, np.ndarray]) -> np.ndarray:  # noqa: ARG002
        """Read the leader and return one action for the follower."""
        leader_obs = self.leader_robot.get_observation()
        action = np.asarray(leader_obs.joint_positions, dtype=np.float32)
        return action[None, :]
```

To do this we change the `InferenceModel` in our runtime and execution classes to be a `RuntimePolicy`, which is a protocol that's implemented by both `InferenceModel` and `TeleoperatorPolicy`.

I'm reopening this mainly to show how teleoperation could be done, though it might be better to spend a bit more time on the interface design for this. Another open that we would need to look into is see if we can add a `RecordingCallback` so that we can record a teleoperation session.


maxxgx:
Closing this as it has been addressed with the generic runtime system

PR #167, opened 2026-06-19
fix: incorrect InferenceModel inputs with one camera setup

sovrasov:
## Summary

- Before fix, single-image policy input contract was broken: policies expect single image under `images` key

## Why

- Bugifx

## Validation

- UTs, policy run

## Breaking changes

- None


alfieroddan (approved):
Thanks!

PR #169, opened 2026-06-22
feat: load InferenceModel from hf hub

sovrasov:
## Summary

- Now `InferenceModel` can be initialized from a hf repo
- `from_pretrained()` mimics the original hf semantic and does not overlap with other initialization paths

## Why

- That eases the way how a model is initialized from a hf repo. Previously, a hf repo should have been downloaded manually in advance

## Validation

```python 
import numpy as np
from physicalai.inference import InferenceModel

if __name__ == "__main__":
    model = InferenceModel.from_pretrained("OpenVINO/act-fp16-ov")
    model.reset()
    observation = {
        "images.image": np.random.rand(1, 3, 256, 256).astype(np.float32),
        "images.image2": np.random.rand(1, 3, 256, 256).astype(np.float32),
        "state": np.zeros((1, 8), dtype=np.float32),
    }
    action = model.select_action(observation)
```

## Breaking changes

- `InferenceModel.load` is deleted


src/physicalai/inference/model.py
```diff
@@ -18,6 +18,7 @@
 from physicalai.inference.data.features import InferenceFeature
 from physicalai.inference.manifest import ComponentSpec, Manifest
 from physicalai.inference.runners import get_runner
+from physicalai.inference.utils._hub import download_from_hub  # noqa: PLC2701
```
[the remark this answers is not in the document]
sovrasov: Already fixed

src/physicalai/inference/model.py
```diff
@@ -179,6 +180,50 @@ def load(
...
+            ... )
+        """
+        local_dir = download_from_hub(
+            repo_id,
+            revision=revision,
+            cache_dir=cache_dir,
+            token=token,
+            allow_patterns=allow_patterns,
+        )
+        return cls(export_dir=local_dir, **kwargs)
```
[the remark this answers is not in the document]
sovrasov: Added

samet-akcay:
thanks for creating this addition. 

one question. what does `load` mean/do? Does it act as `from_checkpoint`, from file? 

if so, would it be an idea if we also have `from_checkpoint` to load from file system. and load can potentially do both depending on the input ?

alfieroddan:
> thanks for creating this addition.
> 
> one question. what does `load` mean/do? Does it act as `from_checkpoint`, from file?
> 
> if so, would it be an idea if we also have `from_checkpoint` to load from file system. and load can potentially do both depending on the input ?

Seems as though we just interact with Huggingface rather than any ckpt or path. Perhaps we could do `.from_hf`?

sovrasov:
When I was adding `from_pretrained` I had exactly the same question about `load` as Samet had: why do we need it? It simply duplicates `__init__` functionality adding nothing on top. If we decide to sort this out, I'd propose to delete `load`.
`from_hf` naming seems clear, but my motivation was to mimic the original hf interface, since it's familiar to users. Have no strong preference here. Typical HF example:
```python
from transformers import AutoModelForCausalLM

model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-hf", device_map="auto")
```
ours:
```python
from physicalai.inference import InferenceModel

model = InferenceModel.from_pretrained("OpenVINO/act-fp16-ov", device="GPU")
```

alfieroddan:
> When I was adding `from_pretrained` I had exactly the same question about `load` as Samet had: why do we need it? It simply duplicates `__init__` functionality adding nothing on top. If we decide to sort this out, I'd propose to delete `load`. `from_hf` naming seems clear, but my motivation was to mimic the original hf interface, since it's familiar to users. Have no strong preference here. Typical HF example:
> 
> ```python
> from transformers import AutoModelForCausalLM
> 
> model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-hf", device_map="auto")
> ```
> 
> ours:
> 
> ```python
> from physicalai.inference import InferenceModel
> model = InferenceModel.from_pretrained("OpenVINO/act-fp16-ov", device="GPU")
> ```

In which case I agree to delete `load` and use `from_pretrained`. Then if `from_pretrained` can also take local dir then that makes sense to me (let's say they did `hf download alfie/model --local-dir alfie-model and used the local dir path).

docs/how-to/inference/load-exported-policy.md
```diff
@@ -15,10 +15,27 @@ model.reset()
...
+```
+
+Pin a revision (branch, tag, or commit SHA) for reproducible loads, and pass a
+token for private repositories.
+
+```python
+model = InferenceModel.from_pretrained(
+    "physical-ai/act-cube",
+    revision="v1.0",
+    token="hf_...",
```
AlexanderBarabanov: In this example we propose user to hardcode HF token that might be not secure - can we use env variable, something like:

```
token = os.environ["HF_TOKEN"]
model = InferenceModel.from_pretrained(
    "physical-ai/act-cube",
    revision="v1.0",
    token=token,
```

src/physicalai/inference/utils/_hub.py
```diff
@@ -0,0 +1,67 @@
...
+    The returned directory mirrors a local export directory and can be passed
+    straight to :class:`~physicalai.inference.model.InferenceModel`.
+
+    Args:
+        repo_id: Hub repository identifier, e.g. ``"physical-ai/act-cube"``.
+        revision: Optional git revision (branch, tag, or commit SHA). Pin to a
+            commit SHA for reproducible, tamper-evident loads.
+        cache_dir: Optional cache directory for the download. Defaults to the
+            standard Hugging Face cache location.
+        token: Optional Hugging Face access token for private repositories.
```
AlexanderBarabanov, line 38: 1) Can we remove `token` and completely rely on token provided in env variable  - as far as I understand `snapshot_download` uses it by default if `token` is not provided. Or there will be some cases where env variable is not enough and we have to explicitly pass the token?

2) If not possible to remove - `snapshot_download` also accept `token = True` - shall we support it as well?

src/physicalai/inference/model.py
```diff
@@ -159,25 +160,65 @@ def chunk_size(self) -> int:
...
+        revision: str | None = None,
+        cache_dir: str | Path | None = None,
+        token: str | None = None,
+        allow_patterns: list[str] | None = None,
         **kwargs: Any,  # noqa: ANN401
     ) -> InferenceModel:
-        """Load inference model with auto-detection.
+        """Load an inference model from a Hugging Face Hub repository or local path.
+
+        If ``repo_id`` points to an existing local directory, it is loaded
```
AlexanderBarabanov, line 174: I propose to add a statement to highlight that consumer shall verify that `repo_id` points to trusted location, something like:
`The consumer shall ensure the path points to a trusted directory. `

Potentially there might be some security issues like path traversal as our code does not check (by design to not reduce function applicability) that `repo_id` does not contain `.` or `..` or malformed `manifest.json`

docs/how-to/inference/load-exported-policy.md
sovrasov: deleted

src/physicalai/inference/utils/_hub.py
sovrasov: deleted

src/physicalai/inference/model.py
sovrasov: Notice added in [02d56a9](https://github.com/openvinotoolkit/physicalai/pull/169/commits/02d56a967d14e08a253f97358e63f36532e53ae4)

docs/how-to/inference/load-exported-policy.md
```diff
@@ -15,10 +15,25 @@ model.reset()
 action = model.select_action(observation)
 ```
 
+Load directly from the Hugging Face Hub with a repo id.
+
+```python
+model = InferenceModel.from_pretrained("physical-ai/act-cube")
```
alfieroddan:
```suggestion
model = InferenceModel.from_pretrained("OpenVINO/act-fp16-ov")
```

alfieroddan (changes requested):
Thanks Vlad!

One minor change for docs :)

alfieroddan (approved):

PR #172, opened 2026-06-26
refactor: speedup int8 resize

sovrasov:
## Summary

- In common resize, input image is not converted to fp32, unless it was already an fp array

## Why

- Most of cameras return uint8 images, and resize of this type is the most effective on CPU

Results on 2x 1920x1080 
Input dtype | Median latency
-- | --
fp32 | 20.728 ms
uint8 | 1.164 ms

</body>
</html>

## Validation

- Unit test that compares fp32 vs int8 resize result

## Breaking changes

- None


alfieroddan (approved):
Thanks!

Only edge case I can think of Depth images may wish to resize in different modes other than `cv2.INTER_LINEAR` but can save for a different PR.

PR #177, opened 2026-06-30
doc: validation proposal

maxxgx:
## Summary

-

## Why

-

## Validation

-

## Breaking changes

- None

## Related issues

- Closes #


docs/design/openvino-validation.md
```diff
@@ -0,0 +1,211 @@
...
+| B70      | Arc Battlemage discrete     |
+
+CPU smoke stays on hosted runners as the portable baseline; GPU/accuracy/perf
+tiers run across this matrix.
+
+## Frequency
+
+| Trigger                      | Scope                                | Gates?       |
+| ---------------------------- | ------------------------------------ | ------------ |
+| OV bump PR / PAI PR          | smoke (CPU)                          | yes          |
```
sovrasov, line 185: I think I'd be useful also figure out model storage approach here. We can have models pre-converted to IRs and stored somewhere (HF?), or obtain them on-the-fly. For some policies export takes significant time (~10min), which would increase CI latency, since we'd like to run smoke tests on every PAI PR

docs/design/openvino-validation.md
```diff
@@ -0,0 +1,211 @@
...
+## Reporting
+
+Per-run record of OV / `openvino_tokenizers` versions, model, platform, tier,
+and pass/fail + metrics. Failures notify the owner; perf/accuracy tracked as a
+trend across versions. Dashboards and notification wiring are OV-side.
+
+## Next
+
+- Build the `ov_smoke` test (three contract stages, model discovery, version logging).
+- Lock `openvino` + `openvino_tokenizers` as a matched pair in
```
AlexanderBarabanov, line 198: We may need to create Renovate rule that will group `openvino` + `openvino_tokenizers` version bump in one PR

docs/design/openvino-validation.md
```diff
@@ -0,0 +1,211 @@
...
+  [`pyproject.toml`](../../pyproject.toml) — they are versioned independently
+  and must be binary-compatible; constraining the tokenizer to the matching
+  minor stops a resolver from installing the mismatch that broke us.
+- Add a small tokenizer-bearing model the gate can run — the tokenizer break
+  only shows on a model with an OV tokenizer; pi05 is 6.25 GB (too large for
+  hosted CI) and ACT has none, so the gate needs a small tokenizer model.
+- Wire `ov_smoke` into [`library.yml`](../../.github/workflows/library.yml) as a
```
AlexanderBarabanov, line 205: Are we going to run `ov_smoke`  on every PR or only for PRs that bump OpenVINO version?
If we need to achieve the second option, it can be implemented via Renovate rule that will open PR with `openvino` + `openvino_tokenizers` bump from pre-defined branch name and then execute  `ov_smoke` if branch_name == predefined.

docs/design/openvino-validation.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Decision:** split by layer.
+
+- **Test logic lives in physicalai** — this is the only way to guarantee its
+  quality and relevancy. Only physicalai knows its real OpenVINO usage, and the
+  test sits next to the code it protects, so it changes in the same PR as that
+  code. The OV team does not author test logic in-house.
+- **Orchestration and reporting live in OpenVINO** — the workflow YAMLs, runners,
+  matrix, and dashboards are OV-side tooling. OV consumes the physicalai test by
+  pinned tag (`git clone physicalai@<tag>` → `pytest -m ov_smoke`).
```
AlexanderBarabanov, line 64: Currently, our tagging strategy covers only [releases](https://github.com/openvinotoolkit/physicalai/tags).
Do we need to maintain PhAI release candidate tags as well?
For weekly tests - can we use `git clone physicalai@main`? In this case we would not need to implement automation on the OpenVINO repo side to periodically bump the version (though it will be mutable).

docs/design/openvino-validation.md
maxxgx: I think second option (on OV bump) will be sufficient to test whether new OV version is compatible

docs/design/openvino-validation.md
maxxgx: Maybe a private HF repo would be more convenient in this case

PR #179, opened 2026-07-02
feat: generic robot runtime

maxxgx:
## Summary

Replaces the policy-only `PolicyRuntime` (which hard-wired inference into the loop and exposed ~25 types: a `Controller` protocol, 4 capability protocols, a lazy `Tick`, a subclass, and `isinstance` dispatch in the loop body) with **one runtime class + one 3-method protocol**. Same fixed-rate loop now drives policy, teleop, and future action sources with no fork and zero `isinstance` in the loop. Also reorganizes the flat `runtime/` package into subpackages.

Scoped to `src/physicalai/runtime/` + its CLI/config surface. No changes to inference/robot/capture layers.

## ActionSource

```python
class ActionSource(Protocol):
    def connect(self, *, bus, session_id) -> None: ...
    def update(self, robot_state, camera_frames, step) -> np.ndarray: ...
    def disconnect(self) -> None: ...
```

`RobotRuntime` owns robot+cameras, timing, resilient IO (retry / stale fallback / circuit breaker), the callback bus, and lifecycle. `PolicySource` and `TeleopSource` are the two concrete implementations. `run(duration_s) -> int` (steps).

## Loop semantics (the core behavioral change)

```mermaid
sequenceDiagram
    participant R as RobotRuntime
    participant S as ActionSource
    participant B as CallbackBus
    participant Rb as Robot

    R->>Rb: read robot_state + camera_frames (once/tick, retry+stale)
    R->>S: update(robot_state, camera_frames, step)
    S-->>R: action
    R->>B: on_action_ready(action) [transform chain]
    B-->>R: action'
    R->>Rb: send_action(action') [resilient]
    R->>B: on_action_sent(action') · emit TickEvent
    Note over R: sleep to hold fps
```

Key points for review: reads are **eager** (once per tick, shared by value — no `Tick`), `update()` **always returns a sendable action** (hold/repeat is the source's internal decision, no runtime `on_hold`), and no `isinstance` anywhere in the loop. Action-source-specific live metrics (queue depth) ride a dedicated `MetricsEvent`, keeping the runtime ignorant of policy concepts.

## Removed / renamed

**Removed:** `PolicyRuntime`, `Tick`, `SupportsBus/Stats/Drain/HoldInfo`, `RunStats`, `on_hold`, shutdown drain, dual config schema.
**Renamed:** `Controller→ActionSource`, `PolicyController→PolicySource`, `TeleopController→TeleopSource`, `before_send_action→on_action_ready`, `runtime.py→core.py`.

## Package structure

Public imports (`from physicalai.runtime import X`) unchanged; only backing files moved.

```text
runtime/
  core.py  events.py  smoothers.py  _callback_bus.py
  action_sources/   # base, policy, teleop
  execution/        # base, sync, async, rtc, queue, rtc_queue
  callbacks/        # console, jsonl, async_callback, rerun, low_pass
  observer/         # telemetry subscriber (+ relocated _telemetry.py, dormant)
```

## Out of scope

- `runtime`→`rollout` naming (deferred). I think rollout is the more established term and it would make sense for us to adopt as well, especially early on before real adoption.
- `observer/_telemetry.py` relocated but dormant (observer is just scaffolding for now).

## Validation

Manually tested teleop via CLI with:

```bash
physicalai run --config examples/runtime/teleop_runtime.yaml
```
`teleop_runtime.yaml`:
```yaml
runtime:
  robot:
    class_path: physicalai.robot.SO101
    init_args:
      port: /dev/ttyACM1
      calibration: /home/max/.local/share/physicalai/robots/b5b8ba26-a470-49a2-ad36-25d69323bb33/calibrations/795d23ad-f3f3-4154-a11d-c46c12985a81.json

  action_source:
    class_path: physicalai.runtime.TeleopSource
    init_args:
      leader:
        class_path: physicalai.robot.SO101
        init_args:
          port: /dev/ttyACM0
          calibration: /home/max/.local/share/physicalai/robots/46a715c3-d07c-49d9-ad0d-290ebe3662e4/calibrations/99bf3448-49a5-4bcd-8f0a-37fa244481ab.json
          role: leader

  fps: 100.0

  callbacks:
    - class_path: physicalai.runtime.ConsoleCallback
      init_args:
        throttle_steps: 100   # print once per second at 100fps

run:
  duration_s: 300
```

## Breaking changes

- Config: `model`/`execution` now nest under an `action_source:` block.
- `PolicyRuntime(...)` → `RobotRuntime(robot, action_source=PolicySource(model, execution), ...)`.
- `run()` returns `int`, not `RunStats`.
- Callbacks: `before_send_action`→`on_action_ready` (return required); read `event.robot_state`/`event.camera_frames` (not `event.tick.*`); drop `on_hold`.

## Related issues

- Closes #


src/physicalai/runtime/action_sources/policy.py
samet-akcay, line 1: would it be an idea to split `controller.py` into a subpackage as controller might/will grow in the future? It's even multiple now? 

so it would be sth like?

```bash
src/physicalai/runtime/controller/
--- controller.py
--- policy.py
--- teleoperation.py
--- ...
```

src/physicalai/cli/run.py
```diff
@@ -53,13 +94,38 @@ def build_parser() -> ArgumentParser:
...
+    Peeks at sys.argv to determine the config schema: if the config file
+    declares ``runtime.class_path``, uses the general (subclass) parser;
+    otherwise uses the flat PolicyRuntime parser for backward compatibility.
+
+    Returns:
+        Parser appropriate for the detected config schema.
+    """
+    if _peek_config_uses_general_schema():
+        return _build_general_parser()
+    return _build_legacy_parser()
```
MarkRedeman: We're only at version 0.1.1, do we need to keep maintaining this separate parser?

src/physicalai/cli/run.py
```diff
@@ -32,18 +33,58 @@
 Runtime constructor arguments are available under --runtime.* when executing
 the command. Use --print_config with a complete command to inspect the full
 jsonargparse schema.
+
+Supports two config schemas:
```
MarkRedeman: Why do we support both? What are the pros / cons of each?

src/physicalai/runtime/controller.py
```diff
@@ -0,0 +1,311 @@
...
+    def holds(self) -> int:
+        """Consecutive hold count from queue pop() behavior."""
+        return self._action_queue.consecutive_holds
+
+    @property
+    def last_was_hold(self) -> bool:
+        """Whether the latest update() returned a held action fallback."""
+        return self._last_was_hold
+
+    def set_bus(self, bus: _CallbackBus, session_id: str) -> None:
```
MarkRedeman: What's the use case of `set_bus` here? If the user is already injecting `_execution` into the `PolicyController`'s constructor then why can't they set the bus directly?

docs/design/components/runtime-system/runtime-architecture.md
MarkRedeman, line 1: General comment: this doc feels very verbose and very AI heavy. I don't have any good concrete feedback yet but at the moment it's hard to read.
I'll probably skip the document in my initial review and focus mainly on the code.

src/physicalai/runtime/callbacks.py
```diff
@@ -137,20 +148,20 @@ def on_tick(self, event: TickEvent) -> None:
...
+                copied = {
                     name: Frame(data=f.data.copy(), timestamp=f.timestamp, sequence=f.sequence)
                     if not f.data.flags.owndata
                     else f
-                    for name, f in event.camera_frames.items()
-                },
-            )
+                    for name, f in camera_frames.items()
+                }
+                event.tick._set_camera_frames(copied)  # noqa: SLF001
```
MarkRedeman: Can we change the method to be public if we assume that it will get accessed publicly?

src/physicalai/runtime/callbacks/rerun.py
```diff
@@ -479,17 +490,26 @@ def _log_camera_frames(self) -> None:
...
+            if self._image_max_dim is not None:
+                data = _downsample_to_max_dim(data, self._image_max_dim)
+
+            img = rr.Image(data)
+            if self._image_jpeg_quality is not None:
+                img = img.compress(jpeg_quality=self._image_jpeg_quality)
+            rr.log(f"camera/{name}", img)
+
+    @staticmethod
+    def _read_camera_frame(sub: Any, name: str) -> Frame | None:  # noqa: ANN401
```
MarkRedeman: Why do we not add a type, `sub: Camera` here?

src/physicalai/runtime/controller.py
```diff
@@ -0,0 +1,311 @@
...
+        return action
+
+    def stop(self) -> None:
+        """Stop execution strategy."""
+        self._execution.stop()
+
+    def drain(self, limit: int) -> Iterator[np.ndarray]:
+        """Yield up to ``limit`` remaining queued actions for shutdown flush.
+
+        Called after :meth:`stop` (execution already quiesced), so no concurrent
```
MarkRedeman:
```suggestion
        Called after :meth:`stop` (execution already paused), so no concurrent
```

I had no clue what `quiesced` meant, thought it was hallucinated before googling  😄 

---

If the assumption is that this is always called after `stop`, then why don't we call it inside of `stop`?

src/physicalai/runtime/action_sources/policy.py
```diff
@@ -0,0 +1,311 @@
...
+            Dictionary ready for model inference.
+        """
+        model_input: dict[str, Any] = {STATE: np.array([robot_obs.state], dtype=np.float32)}
+        image_inputs: dict[str, np.ndarray] = {}
+        # Merge robot-embedded images and external cameras
+        if robot_obs.images:
+            for name, frame in robot_obs.images.items():
+                image_inputs[name] = frame.data[np.newaxis]
+        for name, frame in camera_frames.items():
+            image_inputs[name] = frame.data[np.newaxis]
```
MarkRedeman, line 144: I'm a bit confused about this mergin. What's the use case here? What is the difference between `robot_obs` and `camera_frames`?

src/physicalai/runtime/runtime.py
```diff
@@ -565,8 +523,21 @@ def _resilient_observe(self) -> tuple[RobotObservation, dict[str, Frame]]:
...
 
-        return robot_obs, camera_frames
+    def _resilient_observe(self) -> tuple[RobotObservation, dict[str, Frame]]:
+        """Read robot observation and camera frames with retry and stale fallback.
+
+        Returns:
+            Tuple ``(robot_observation, camera_frames)``.
+        """
+        tick = Tick(
+            frame_index=0,
```
MarkRedeman: Why is `frame_index` always 1 here?
It feels a bit weird that we are constructing a tick but then don't return it.

src/physicalai/runtime/runtime.py
```diff
@@ -663,9 +637,57 @@ def _shutdown(self, step: int) -> None:
...
+    def _action_queue(self) -> ActionQueue:
+        return self._policy_controller.action_queue
+
+    def _build_model_input(self) -> dict[str, Any]:
+        robot_obs = self._robot.get_observation()
+        camera_frames = {name: cam.read_latest() for name, cam in self._cameras.items()}
+        return self._build_model_input_from(robot_obs, camera_frames)
+
+    def _build_model_input_from(self, robot_obs: RobotObservation, camera_frames: dict[str, Frame]) -> dict[str, Any]:
+        return self._policy_controller.to_model_input(robot_obs, camera_frames)
```
MarkRedeman: Who is using these methods/properties? They are marked as private but I don't see them being used inside of this class.

MarkRedeman (commented):
Some comments. Didn't end up fully reviewing the actual design doc yet.

Generally this looks like a nice addition, but I am concerned that this feels overly complex. Keeping a mental model of which components get used where is not easy.
Maybe we can add some diagrams that show how the different runtimes can be used and where each code gets executed (i.e. in a separate thread, process etc?)

docs/design/components/runtime-system/runtime-architecture.md
maxxgx: Please feel free to skip this doc, as it's not meant to be shipped and it's only included while the PR is in draft state. It mainly serves as an implementation reference for agents. 

After the implementation is finalized, we'll remove it and update the API docs instead.

src/physicalai/runtime/action_sources/policy.py
maxxgx: Yeah current structure is like this

src/physicalai/cli/run.py
maxxgx: no dual schema anymore

src/physicalai/runtime/callbacks/rerun.py
maxxgx: removed separate cameras reading from rerun. The runtime will emit an event with the observation and rerun will log that. This is simpler and also avoids frame discrepancies between observation and logging.

src/physicalai/runtime/action_sources/policy.py
maxxgx: As you may remember, early on we decided to separate cameras from robots but at the same time allow robots with cameras. As a result we `image_inputs` can be contained both inside `robot_obs` or separate `camera_frames`. Since they named, merging should be safe

src/physicalai/runtime/runtime.py
maxxgx: Tick has been removed

src/physicalai/runtime/runtime.py
maxxgx: removed, as this is not runtime responsibility.

maxxgx:
@MarkRedeman @samet-akcay I've redesigned the runtime architecture, with the priority of simplifying it while retaining the capabilities. Please check the updated PR description for an overview

docs/reference/runtime-api.md
```diff
@@ -21,14 +19,50 @@ The most important methods are shown below.
 ```python
 runtime.connect() -> None
 runtime.disconnect() -> None
-runtime.run(duration_s: float | None = None) -> RunStats
+runtime.run(*, duration_s: float | None = None) -> int
+```
+
+`RobotRuntime` also supports context-manager usage so connections are cleaned up automatically. `run()` returns the number of steps completed this run — there is no aggregate stats object. Other stats are read directly off the objects the caller already holds, e.g. `runtime.action_source.action_queue.total_pops` or `execution.inference_count`.
```
MarkRedeman: Might be good to elaborate what `steps` means here. I assume it is a single loop where you set send an action from an action source.
Others might also interpret a step as a single episode?

src/physicalai/runtime/action_sources/policy.py
MarkRedeman: Ah right I forgot about that. I feel like it might make things a bit more complicated. I'd personally prefer to remove the cameras from the robot class so that the code can be simpler. But that's out of scope here, let's check back on that another time.

src/physicalai/runtime/action_sources/policy.py
```diff
@@ -0,0 +1,162 @@
...
+
+        if len(image_inputs) > 1:
+            for name, data in image_inputs.items():
+                model_input[f"{IMAGES}.{name}"] = data
+        elif len(image_inputs) == 1:
+            model_input[IMAGES] = next(iter(image_inputs.values()))
```
MarkRedeman, line 150: Why do we ignore the name of the image input if there's only 1? If I were to record with only an `overview` camera then I think I'd still expect the model to use its name as an input. Mainly so that it would feel more consistent with when retraining with 2 cameras?

(No need to address in this PR.)

src/physicalai/runtime/callbacks/console.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from physicalai.runtime.events import LifecycleEvent, TickEvent
+
+
+class ConsoleCallback:
+    """Periodic one-line summary to stdout (~1 per second)."""
+
+    def __init__(self, throttle_steps: int = 30) -> None:  # noqa: D107
```
MarkRedeman: I guess `~1 second` here depends on `throttle_steps` and the actual fps of the runtime?

src/physicalai/runtime/_callback_bus.py
```diff
@@ -60,18 +60,29 @@ def emit_lifecycle(self, event: LifecycleEvent) -> None:
             except Exception:
                 logger.exception("Callback %r failed in on_lifecycle", cb)
 
-    def invoke_before_send_action(self, *, action: np.ndarray, step: int) -> np.ndarray:
+    def invoke_on_action_ready(self, *, action: np.ndarray, step: int) -> np.ndarray:
+        """Chain each callback's ``on_action_ready``, threading the return value.
+
+        Every callback must return a valid action (no ``None`` sentinel) — a
+        callback that doesn't want to change anything returns its input
+        unchanged.
```
MarkRedeman, line 68: Is the input of each callback the output of the previous callback, or do they all get the output from the action source?

Typically with these things I quite like having a middleware approach:
```
def my_callback(input, next)
    new_input = self.do_something_with_input(input)
    output = next(new_input)
    new_output = self.do_something_with_output(output)

    return new_output
```

so then each callback would be able to both tweak the input and output.
I don't know if that's something that would be directly useful here though. The one example I was thinking of is that you could use `LowPassFilterCallback` and have it make sure the output is always smooth, but even in the middleware case order of your callbacks would still matter.

src/physicalai/runtime/core.py
```diff
@@ -0,0 +1,438 @@
...
+logger = logging.getLogger(__name__)
+
+
+_MAX_OBS_RETRIES = 3
+_MAX_SEND_RETRIES = 2
+_RETRY_BACKOFF_S = 0.001
+_GOAL_TIME_TICKS = 3
+
+
+class RuntimeCallback(Protocol):
```
MarkRedeman, line 39: Perhaps move this to `_callback_bus.py`?

src/physicalai/runtime/core.py
```diff
@@ -0,0 +1,438 @@
...
+        self._action_source = action_source
+        self._fps = fps
+        self._cameras: Mapping[str, Camera] = cameras or {}
+        self._bus = _CallbackBus(callbacks)
+        self._goal_time = (1.0 / fps) * _GOAL_TIME_TICKS
+        self._connected = False
+        self._last_robot_obs: RobotObservation | None = None
+        self._last_camera_frames: dict[str, Frame] = {}
+        self._consecutive_error_ticks: int = 0
+        self._max_consecutive_error_ticks: int = int(3 * fps)
```
MarkRedeman, line 89: Might be interesting to apply a [CircuitBreaker pattern](https://martinfowler.com/bliki/CircuitBreaker.html) here. We could wrap the robot and camera instances with a small decorator that implements that pattern.

MarkRedeman (dismissed):

docs/reference/runtime-api.md
maxxgx: Added definition

src/physicalai/runtime/callbacks/console.py
maxxgx: yeah default target is 1s. Updated docstring

src/physicalai/runtime/core.py
maxxgx: I agree, created an issue to track this: https://github.com/openvinotoolkit/physicalai/issues/186

MarkRedeman (approved):

src/physicalai/runtime/core.py
```diff
@@ -0,0 +1,440 @@
...
+
+        for attempt in range(_MAX_SEND_RETRIES):
+            try:
+                self._robot.send_action(action, goal_time=self._goal_time)
+            except (ConnectionError, OSError) as exc:
+                last_error = exc
+                if attempt + 1 < _MAX_SEND_RETRIES:
+                    time.sleep(_RETRY_BACKOFF_S)
+            else:
+                self._consecutive_error_ticks = 0
```
AlexanderBarabanov, line 390: As `_consecutive_error_ticks` is used by both `_read_robot_resilient()` and `_resilient_send()`,  I am thinking about a case when `_read_robot_resilient()` fails and sets `_consecutive_error_ticks += 1` but then `_resilient_send()` is ok and it sets `_consecutive_error_ticks -> 0` (in the same while loop).
In this case, counter can never accumulate toward `_max_consecutive_error_ticks`, although `_read_robot_resilient()` always fails.

Is this a real scenario? If real, should we separate the read failure counter from the send failure counter?

src/physicalai/runtime/core.py
```diff
@@ -0,0 +1,440 @@
...
+        self._last_tick_stale = stale
+        camera_frames = self._read_cameras_resilient()
+        return robot_state, camera_frames
+
+    def _resilient_send(self, action: np.ndarray) -> None:
+        last_error: ConnectionError | OSError | None = None
+
+        for attempt in range(_MAX_SEND_RETRIES):
+            try:
+                self._robot.send_action(action, goal_time=self._goal_time)
```
AlexanderBarabanov, line 384: Should we validate provided action for some allowed range or at least  with `np.isfinite(action)`?

src/physicalai/runtime/_callback_bus.py
```diff
@@ -60,18 +60,29 @@ def emit_lifecycle(self, event: LifecycleEvent) -> None:
...
             try:
-                modified = fn(action=result, step=step)
-                if modified is not None:
-                    result = modified
+                modified_action = fn(action=result, step=step)
+                if modified_action is None:
+                    logger.warning("Callback %r returned None from on_action_ready, ignoring", cb)
+                else:
+                    result = modified_action
             except Exception:
```
AlexanderBarabanov, line 84: As far as i understand, in case of exception, the runtime continues with a log message without clear runtime signal - should we somehow propagate exception case to the consumer code to let them know that there is an issue?
Perhaps, we could just remove try/except block here to propagate exception.

src/physicalai/runtime/action_sources/teleop.py
```diff
@@ -0,0 +1,67 @@
...
+            self._leader.connect()
+            self._leader_owned = True
+
+    def update(self, robot_state: RobotObservation, camera_frames: Mapping[str, Frame], step: int) -> np.ndarray:  # noqa: ARG002
+        """Read the leader arm and return the action for the follower.
+
+        Returns:
+            Action array for the follower robot.
+        """
+        return self._to_action(self._leader.get_observation())
```
AlexanderBarabanov, line 61: Are there any potential issues (like ConnectionError, OSError) that can be originated from `._leader.get_observation`? Perhaps we could use try/catch as we do for `_read_robot_resilient()` or `_resilient_send()`

src/physicalai/runtime/core.py
```diff
@@ -0,0 +1,440 @@
...
+            logger.exception("Worker died during runtime")
+            raise
+        finally:
+            self._shutdown(step)
+
+        return step
+
+    def _reset_session(self) -> None:
+        """Reset all session-scoped state for a fresh run."""
+        self._session_id = uuid.uuid4().hex[:8]
```
AlexanderBarabanov, line 270: What is the purpose of this `_session_id `? Perhaps we may add a small comment to explain that this sesion_id is not for security but for telemetry correlation purposes.

src/physicalai/runtime/action_sources/teleop.py
```diff
@@ -0,0 +1,67 @@
...
+        self._to_action = to_action or (lambda obs: obs.joint_positions)
+        self._leader_owned = False
+
+    def connect(self, *, bus: _CallbackBus, session_id: str) -> None:  # noqa: ARG002
+        """Connect leader if not already connected."""
+        if not self._leader.is_connected():
+            self._leader.connect()
+            self._leader_owned = True
+
+    def update(self, robot_state: RobotObservation, camera_frames: Mapping[str, Frame], step: int) -> np.ndarray:  # noqa: ARG002
```
AlexanderBarabanov, line 55: Should we validate a value against some allowed range or with `np.isfinite(action)` before return?

AlexanderBarabanov (approved):
Thanks, comments can be addressed (if required) in a separate PR.

src/physicalai/runtime/core.py
maxxgx: I think for this to happen in practice, we'd need the first read to be successful so that `self._last_robot_obs` is set, and all subsequent reads to fail, since otherwise a `ConnectionError` is raised immediately.

It's a real but very unlikely edge case, as write and read go through the same wire, so they should either both succeed or fail.

Will consider this for https://github.com/openvinotoolkit/physicalai/issues/186

src/physicalai/runtime/core.py
maxxgx: I think this is robot driver's responsibility. Robots should know its own allowed range

src/physicalai/runtime/_callback_bus.py
maxxgx: Good point, will keep the exception log and reraise it (in follow up PR)

src/physicalai/runtime/action_sources/teleop.py
maxxgx: Yeah that could be an issue. However, I'd wait before adding the same mechanism here as we may have a better solution soon after https://github.com/openvinotoolkit/physicalai/issues/180 or https://github.com/openvinotoolkit/physicalai/issues/186

src/physicalai/runtime/core.py
maxxgx: Will add in the follow up

maxxgx:
@AlexanderBarabanov, thanks for the review. I'll create another PR to address some of your comments

PR #184, opened 2026-07-07
fix: OpenVINO optimization notebook deployment flow in 003 notebook

zhuo-yoyowz:
## Summary
Fixed issues in 003_OpenVINO_Optimization.ipynb, including:
1) Fixed git clone Physical AI commit number, since the repo is developed fast and the last version used API out of date, causing errors during running.
2) Removed the hard-coded robot calibration file path. Remind users to manually config the file path according to their env.
3) Added a reminder for user to give permission to USB cameras ports in Linux OS.
4) Added two choices for Model Loading, either using exported model already in local env or downloading an open-sourced pi-0.5 fine-tuned model. 
5) Updated some APIs according to the updated version of Physical API master branch.


examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -49,7 +49,9 @@
    "id": "3c28a826-a284-4623-8a7b-74765529bdd6",
    "metadata": {},
    "source": [
-    "Clone the OpenVINO Physical AI repo and install."
+    "Clone the OpenVINO Physical AI repo and install a pinned revision.\n",
+    "\n",
+    "> **Why pin the revision?** The OpenVINO Physical AI APIs are evolving quickly. This notebook depends on a tested `physicalai` runtime revision, so the install cell below checks out a known-good commit before installing the package in editable mode. Update `PHYSICALAI_COMMIT` only after re-validating the rest of this notebook."
```
maxxgx: I think these tutorial were written before release. Now there is no need to clone the physicalai repo since it's already available on pypi, so you could pin a released version

maxxgx (commented):
Thanks for the PR. 

Do you find these tutorials helpful? I was thinking we could reduce them, as they have quite a bit of overlap. For example, the `collect_train_deploy` notebook covers most of the topics.

Also, I believe it will be hard to keep this branch up to date, since breaking changes are quite regular while we're still in early development.

zhuo-yoyowz:
> Thanks for the PR.
> 
> Do you find these tutorials helpful? I was thinking we could reduce them, as they have quite a bit of overlap. For example, the `collect_train_deploy` notebook covers most of the topics.
> 
> Also, I believe it will be hard to keep this branch up to date, since breaking changes are quite regular while we're still in early development.

Thank you for your comments!
We do think Jupyter notebooks are helpful to developers, especially when they try to get hands-on to a whole new set of tools. 
I tested all notebooks. Of course, 001 and 002 are documentations actually, therefore, in the future, maybe we could move them to Readme, so that developers could read them directly when navigating to the "tutorials" path?
004 notebook provides good pure simulation sample codes, which I think is also critical, especially to developers who currently don't own any physical robotic arms to test, but want to see how's everything working, whether it's hard to run and some initial performance results.
003 shows good optimization and deployment codes, but need some fixes, which I think could be also helpful when developers download an open-sourced model and would like to connect the robotic arms to run some initial inferences.
005`collect_train_deploy` does contain a full e2e process, however, it needs at least an A100 to run from the beginning, i.e. the model training. If only keeps this one notebook, developers could run nothing but have to leave the webpage soon if they don't own an A100.
So I hope these notebooks could be merged to the master branch, and we could provide clear guidance on how to use these notebooks to get started quickly and easily. I'm happy to be part of this for maintaining and upgrading, also writing tech blogs to promote these.

examples/tutorials/003_OpenVINO_Optimization.ipynb
zhuo-yoyowz: Yes, just found the formal release version. Will modify the commit number to the released version.

maxxgx:
> > Thanks for the PR.
> > Do you find these tutorials helpful? I was thinking we could reduce them, as they have quite a bit of overlap. For example, the `collect_train_deploy` notebook covers most of the topics.
> > Also, I believe it will be hard to keep this branch up to date, since breaking changes are quite regular while we're still in early development.
> 
> Thank you for your comments! We do think Jupyter notebooks are helpful to developers, especially when they try to get hands-on to a whole new set of tools. I tested all notebooks. Of course, 001 and 002 are documentations actually, therefore, in the future, maybe we could move them to Readme, so that developers could read them directly when navigating to the "tutorials" path? 004 notebook provides good pure simulation sample codes, which I think is also critical, especially to developers who currently don't own any physical robotic arms to test, but want to see how's everything working, whether it's hard to run and some initial performance results. 003 shows good optimization and deployment codes, but need some fixes, which I think could be also helpful when developers download an open-sourced model and would like to connect the robotic arms to run some initial inferences. 005`collect_train_deploy` does contain a full e2e process, however, it needs at least an A100 to run from the beginning, i.e. the model training. If only keeps this one notebook, developers could run nothing but have to leave the webpage soon if they don't own an A100. So I hope these notebooks could be merged to the master branch, and we could provide clear guidance on how to use these notebooks to get started quickly and easily. I'm happy to be part of this for maintaining and upgrading, also writing tech blogs to promote these.

Thank you for your feedback. I think we still improvements before merging it to main branch:
- 002 can be removed and we should point users to the PhysicalAI Studio documentation since it has released an overview of the application: https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/README.md
- I find the code in 003 and 004 not "clean" and often deviates from the recommended usage shown in the repo. It looks like that several shortcuts are used to make the code work. 
- The hugging face links points to personal repos, we can use the official collection instead: https://huggingface.co/collections/OpenVINO/physical-ai
- If would be amazing if we could use a proper simulation environment like mujoco running on the browser: https://lerobot.binhph.am/

maxxgx (approved):

zhuo-yoyowz:
> > > Thanks for the PR.
> > > Do you find these tutorials helpful? I was thinking we could reduce them, as they have quite a bit of overlap. For example, the `collect_train_deploy` notebook covers most of the topics.
> > > Also, I believe it will be hard to keep this branch up to date, since breaking changes are quite regular while we're still in early development.
> > 
> > 
> > Thank you for your comments! We do think Jupyter notebooks are helpful to developers, especially when they try to get hands-on to a whole new set of tools. I tested all notebooks. Of course, 001 and 002 are documentations actually, therefore, in the future, maybe we could move them to Readme, so that developers could read them directly when navigating to the "tutorials" path? 004 notebook provides good pure simulation sample codes, which I think is also critical, especially to developers who currently don't own any physical robotic arms to test, but want to see how's everything working, whether it's hard to run and some initial performance results. 003 shows good optimization and deployment codes, but need some fixes, which I think could be also helpful when developers download an open-sourced model and would like to connect the robotic arms to run some initial inferences. 005`collect_train_deploy` does contain a full e2e process, however, it needs at least an A100 to run from the beginning, i.e. the model training. If only keeps this one notebook, developers could run nothing but have to leave the webpage soon if they don't own an A100. So I hope these notebooks could be merged to the master branch, and we could provide clear guidance on how to use these notebooks to get started quickly and easily. I'm happy to be part of this for maintaining and upgrading, also writing tech blogs to promote these.
> 
> Thank you for your feedback. I think we still improvements before merging it to main branch:
> 
> * 002 can be removed and we should point users to the PhysicalAI Studio documentation since it has released an overview of the application: https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/README.md
> * I find the code in 003 and 004 not "clean" and often deviates from the recommended usage shown in the repo. It looks like that several shortcuts are used to make the code work.
> * The hugging face links points to personal repos, we can use the official collection instead: https://huggingface.co/collections/OpenVINO/physical-ai
> * If would be amazing if we could use a proper simulation environment like mujoco running on the browser: https://lerobot.binhph.am/

Thanks for the feedback. 

I updated 003 to use the released PyPI package flow end-to-end instead of relying on a cloned local physicalai repo. The hardware extras are now installed from the pinned release as well.

I also replaced the personal Hugging Face repos with the official OpenVINO Physical AI collection model (`OpenVINO/pi05-libero-fp16-ov`) and added a guard to make clear that the official hub model is for model-loading validation unless the user verifies that its robot/camera/action schema matches their real SO101 setup.

For the broader 004 cleanup and browser-based simulation direction, I'll submit new PR to fix the issues. For the 002 notebook suggestion, yes, I think we could remove it and point users to documentation as you suggested. But for now, I think we need to show the tutorials repo first and could improve it step by step, so maybe this notebook could be removed later?

maxxgx:
I've looked at the current state of the 003 notebook and it'll need a substantial refactoring but I think it's managable (you can merge this PR first and improve later).

- Since `0.2.0` is not released yet, we can install from the repo. But please remove cell 7 and  11, and replace with a single command, see [how including extras](https://stackoverflow.com/questions/30239152/specify-extras-require-with-pip-install-e). 
- Step 1 "Train a model using...": unnecessary and can be removed, you can refer to PAI Studio path in the model section.
- By default require the download of the model from HF. We are adding exportable runtime config to PAI Studio, so its users can run a model via CLI and, therefore, likely won't be the audience of this notebook.
- Please revise cell 13, it's a big block of (mostly) unrelated code that's quite distracting.
- Camera discovery and selection: you can consider the function `select_cameras_interactive` from camera.factory.
- There is a lof of comments inside code cells. I think it should be minimized since md cells are a better fit vs long comments.
- Calibration paths have changed in Studio. Do not assume the path. If user comes from Studio, they can now download the calibration file directly.
- Should **never** call private methods such as `_build_model_input` in public tutorials.
- `with runtime:` context manager already disconnect all hardware on exit.

I've asked an agent to generate outdated API calls in the notebooks, I think it should be a good starting point:
Notebook calls (0.1.1) | Status | Replacement on main
-- | -- | --
PolicyRuntime(robot=, model=, execution=, cameras=, task=) | Removed | RobotRuntime(robot=, action_source=PolicySource(model=, execution=, task=), fps=, cameras=)
InferenceModel.load(dir, ...) | Removed | InferenceModel(dir, ...) directly, or InferenceModel.from_pretrained("OpenVINO/…") for the Hub path
SharedCamera("uvc", device=…, width=…) | Signature changed | create_camera("uvc", shared=True, device=…, width=…) — or SharedCamera(camera={"class_path":…,"init_args":…})
stats = runtime.run(…) then stats.steps | Type changed | run() now returns int (step count). RunStats is gone; hold/inference counts come from callbacks or the queue.
runtime._build_model_input() | Removed | No replacement — it was private. Validate camera count against model.input_features instead.
SyncExecution(fps=FPS, request_threshold=…) | Never valid | SyncExecution(request_threshold=…). fps belongs to the runtime. This was already wrong in 0.1.1.
ActionQueue() | Not instantiable | It is a Protocol. Use ChunkedActionQueue(...), or omit — PolicySource defaults to one.
snapshot_download(local_dir_use_symlinks=False) | Removed upstream | Drop the kwarg. Gone in huggingface_hub 1.x; the repo allows <2.0, so it is installable and broken.
from physicalai.capture import discover_all | Unchanged | Still exported. Safe.
SO101(port=, calibration=, role="follower") | Unchanged | Still valid. Now also accepts a dict and has an allow_uncalibrated escape hatch.


Unrelated to 003: 
- I think we agree on removing 001 and 002 and put the useful info in the readme. 
- Let's remove the whole `tutorials/media/` folder. Personally I'm not a fan of the images/gifs present there. But most importantly, we cannot commit those in main branch as it would increate the git history size by lot every time they are changed, making the repo bloated.

zhuo-yoyowz:
Thanks for the detailed feedback. Commit b889615 addresses the above comments for notebook 003:

- Consolidated installation into one command with the required SO101 and transport extras.

- Made the official Hugging Face model the default loading path.

- Documented both Studio-exported and externally exported local model workflows.

- Updated inference to the current public APIs: InferenceModel, InferenceModel.from_pretrained, RobotRuntime, PolicySource, and SyncExecution.

- Replaced custom camera handling with select_cameras_interactive and added an optional manual path using create_camera.

- Added camera metadata fallback support for older external model packages.

- Added Studio-downloaded and custom calibration paths without assuming a Studio cache location.

- Removed private API calls and manual hardware cleanup; the runtime context manager now owns connection and cleanup.

- Reduced code-cell comments and moved workflow guidance into Markdown cells.

- Added compatibility checks for camera names and SO101 state/action dimensions.

The updated notebook was also validated end-to-end with a local fine-tuned model, an SO101 robot, and two cameras. It completed 1,800 control steps with no action queue holds.

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
+    "   - **External training and export:** train on another platform, such as a CUDA system, then export a complete OpenVINO policy package yourself. Keep the model IR, manifest, preprocessing/postprocessing assets, normalization statistics, and feature schema together. A standalone `.xml`/`.bin` pair may not contain enough deployment metadata.\n",
     "\n",
-    "%pip install -q --extra-index-url https://download.pytorch.org/whl/cpu -r {requirements_file}\n",
-    "%pip install -q -e physicalai"
+    "The official Pi0.5 checkpoint is trained for the LIBERO benchmark. Its 8-dimensional state and 7-dimensional action interface are not compatible with a six-motor SO101, so keep `RUN_ON_ROBOT = False` when using this default model.\n",
+    "\n",
+    "See the [Physical AI Studio documentation](https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/README.md) for its model and hardware workflows."
```
maxxgx, line 55: I agree that we should not run with the LIBERO trained model, users might think our software doesn't work if we just move the robot randomly.

If the goal of this notebook is to run a working Pi05 model on a real SO101 robot as fast as possible without user training it, then I think we need to provide another model and guidelines on the physical robot setup and task, for example this tutorial does a great job at that https://docs.nvidia.com/learning/physical-ai/sim-to-real-so-101/latest/index.html.

However, if the notebook is just a tutorial on how to run a model within `physicalai`, then I'd redirect the user to train their own model first (with PAI Studio) and then plug in the inputs here. This keeps the notebook's scope more limited and focused.

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
    "source": [
-    "Import libraries, and set up a working directory"
+    "### Action required: Select and configure a model source\n",
+    "\n",
+    "Keep `MODEL_SOURCE = \"huggingface\"` for the default loading validation, or set it to `\"local\"` for an extracted Studio export or an externally exported package.\n",
+    "\n",
+    "- For `\"huggingface\"`, keep the official model ID or set `HUGGINGFACE_MODEL_ID` to your published policy package.\n",
+    "- For `\"local\"`, set `LOCAL_MODEL_PATH` to the root of the extracted OpenVINO policy package.\n",
+    "- Set `TASK` to the instruction used by a language-conditioned policy, or `None` for a policy that does not use language.\n",
+    "- Current Studio exports should declare feature metadata. For an older or external package that does not, set `FALLBACK_CAMERA_NAMES` to the exact camera feature names used during data collection and training.\n",
```
maxxgx, line 70: What do you mean by "older or external package"? 
We have only released one version: 0.1.0 -> https://pypi.org/project/physicalai-studio/

There is no need to support pre-release versions

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
    "source": [
-    "<br>Make sure to copy all the model files from the studio to the $EXPORT_DIR\n",
-    "<br>Then, load the model to the target device\n",
-    "<br><br>\n",
-    "<img src=\"media/download-model.png\" alt=\"Description\" width=\"800\" height=\"200\">"
+    "## 4) Validate Deployment Compatibility\n",
+    "\n",
+    "The runtime maps robot observations and named camera frames to the public feature schema in `policy.input_features`. The model state and action dimensions must match the robot joint count. Multi-camera names must match the suffixes of the model's visual feature names.\n",
+    "\n",
+    "This validation uses only public APIs; it does not construct a private runtime input."
```
maxxgx, line 260: This sentence seems to be aimed at the PR reviewer 😃

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
+    "cameras = {}\n",
+    "if RUN_ON_ROBOT:\n",
+    "    if CAMERA_SOURCE == \"interactive\":\n",
+    "        cameras = select_cameras_interactive(\n",
+    "            width=CAMERA_WIDTH,\n",
+    "            height=CAMERA_HEIGHT,\n",
+    "            fps=CAMERA_FPS,\n",
+    "        )\n",
+    "    elif CAMERA_SOURCE == \"manual\":\n",
+    "        for config in MANUAL_CAMERA_CONFIGS:\n",
```
maxxgx, line 346: I'd go with one of these options, not both. `select_cameras_interactive` keeps things simple. Proper camera configuration could be a separate notebook

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
-    "for name, cam in cameras.items():\n",
-    "    cam.disconnect()\n",
-    "    print(f\"Camera '{name}' disconnected\")\n",
+    "from physicalai.runtime import PolicySource, RobotRuntime, SyncExecution\n",
+    "\n",
+    "if RUN_ON_ROBOT:\n",
+    "    execution = SyncExecution(request_threshold=0.5)\n",
```
maxxgx, line 395: We recently ran a few experiments and sync inference actually performs much better with `request_threshold=0.0` which uses the whole action chunk instead of half (0.5).

examples/tutorials/003_OpenVINO_Optimization.ipynb
```diff
@@ -2,388 +2,417 @@
...
-   "metadata": {},
-   "source": [
-    "## 5) Run the model (policy) on the robot"
+    "\n",
+    "visual_features = features_of_type(policy.input_features, \"VISUAL\")\n",
+    "state_features = features_of_type(policy.input_features, \"STATE\")\n",
+    "action_features = features_of_type(policy.output_features, \"ACTION\")\n",
+    "\n",
+    "declared_camera_names = [\n",
+    "    feature.name.removeprefix(\"images.\")\n",
```
maxxgx, line 279: `PolicySource` only uses `images.<name>` keys when there is more than one image.
With a single camera it puts the frame under the bare `images` key and drops the
name entirely (policy.py::L148):

```python
if len(image_inputs) > 1:
    for name, data in image_inputs.items():
        model_input[f"{IMAGES}.{name}"] = data
elif len(image_inputs) == 1:
    model_input[IMAGES] = next(iter(image_inputs.values()))

maxxgx (commented):
Thanks for the improvements, now the notebook is in a much better shape.

I left a few more comments. I think this could be a great notebook if it was more
focused and deliberate. `physicalai` runs a policy on a robot in 17 lines, see
docs/getting-started/run-a-policy.md. This notebook is 180 lines of code. I think
it can be a lot shorter.

The issue with offering multiple paths is that the reader can't tell which lines
are `physicalai` and which are notebook plumbing, so the API doesn't stand out.
Every reader pays for options that only a few will use. Those alternatives read
better as a sentence in md, or as a link to the docs. Same for the md cells, I
think they can be shorter, and some of the content belongs in the docs instead.

My main concern is that with the default `RUN_ON_ROBOT = False`, "Run All" loads
the model and then prints "skipped" 5 times. It never runs inference, so someone
trying the library for the first time sees nothing actually work. I'd like the
no-hardware path to do something real, then the SO101 part can be the cells at
the end that need hardware. This is in the openvino_notebooks best practices,
notebooks should run in under 5 min with "Run All".
https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/CONTRIBUTING.md

Concretely I would remove:

- `MODEL_SOURCE`, `CALIBRATION_SOURCE` and `CAMERA_SOURCE`. Pick the recommended
  path in code and mention the alternative in md.
- `FALLBACK_CAMERA_NAMES`. Current exports declare feature metadata, we shouldn't
  work around older ones in a tutorial.
- the `RUN_ON_ROBOT` branches.
- the compatibility check cell, see my inline comment. With a single camera
  `PolicySource` puts the frame under the bare `images` key and drops the camera
  name entirely, so comparing names against the `images.` suffixes rejects setups
  the runtime would have accepted. I'd rather have this check in `RobotRuntime`
  than in the notebook.

To be clear, I'm not asking to remove the config cell. One cell with the values a
user has to edit is good, it's the branch selectors I'd like gone.

PR #189, opened 2026-07-13
feat: robot transport via zenoh

### LINKED ISSUE #180: feat: shared robot transport (pub-sub)
### Problem statement

Every robot consumer talks to the serial wire directly (via `get_observation()` / `send_action()`). These calls are blocking, which creates two issues:

- Latency: blocking reads/writes stall the inference loop while waiting on the bus. Multi-robot doesn't scale as latency adds up (e.g. left arm moves first than right arm).
- Redundant bus traffic: with multiple consumers (inference, recording, logging), each hits the serial bus separately for the same data.

### Proposed solution

Apply the SharedCamera model to robots: one owner process per bus, sharing state over iceoryx2 (or zenoh).

- Reads (fan-out): the owner reads state and publishes to shared memory. Consumers subscribe, they never touch the wire, so adding consumers adds no bus traffic.
- Writes (fan-in): producers publish desired actions to a command channel; the owner applies them. For now,  last-writer-wins should be sufficient

Read path is the same as `SharedCamera`; the new work is the write path.

### Alternatives considered

- Background reader thread only and no register: isolates read jitter but doesn't solve multiple consumers or address writes.
- `asyncio` instead of threads: serial libraries are blocking; threads are simpler and I/O already releases the GIL, so async does not offer improvements over threads.

### Additional context

_No response_

maxxgx:
## Summary

Adds Zenoh transport for sharing a physical robot between processes and over network.

```mermaid
flowchart LR
    subgraph Owner["Owner process"]
        HW["Robot driver"] <--> LOOP["Single write-first loop"]
    end
    subgraph Clients["Subscriber processes"]
        A["SharedRobot A"]
        B["SharedRobot B"]
    end

    LOOP -- "publish" --> STATE["/{name}/state"]
    ACTION["/{name}/action"] -- "latest action" --> LOOP
    LOOP -. "answer queries" .-> META["/{name}/metadata"]
    STATE --> A
    STATE --> B
    A --> ACTION
    A -. "probe/validate" .-> META
    B -. "probe/validate" .-> META
```

- One owner exclusively connects to the robot.
- `SharedRobot` instances create-or-attach by name.
- Host-local name and device locks prevent ownership conflicts.
- Local-only loopback is the default; remote mode is explicit.
- State, actions, and metadata use low-latency, latest-wins Zenoh channels.

The [design doc](https://github.com/openvinotoolkit/physicalai/blob/docs/design/docs/design/components/robot-zenoh-transport.md) contains the full architecture. Please note that the design will be moved to the `docs/design` branch before this PR is merged (moved).

#### 1. Teleop tick time: SharedRobot vs. direct driver

Single teleoperation runtime loop with one leader and one follower SO101:

| Config | n ticks | mean | p50 | p99 |
|---|---|---|---|---|
| Direct `SO101` driver | 14,244 | 2.53ms | 2.52ms | 2.77ms |
| `SharedRobot` | 11,771 | 0.108ms | 0.098ms | 0.321ms |

#### 2. Pure Zenoh latency (no owner tick-loop)

| Method | n | mean | p50 | p99 | max |
|---|---|---|---|---|---|
| same-process pub/sub, 64B payload, busy-poll | 1000 | 173.2us | 43.7us | 760.3us | 1071.2us |
| IPC query round-trip vs. real owner subprocess | 200 | 0.32ms | 0.24ms | 1.33ms | 1.59ms |

#### 3. Cross-host Zenoh latency (RTT via query/reply)

Round-trip time measured on the requester's clock only (query out → reply back); one-way estimate is RTT/2. Host A connected via LAN and host B via WLAN on same router.

| Method | n | RTT mean | RTT p50 | RTT p99 | RTT max | one-way estimate (mean/2) |
|---|---|---|---|---|---|---|
| Cross-host query/reply round trip | 500 | 4.274ms | 4.026ms | 6.693ms | 8.166ms | 2.137ms |
| `ping` baseline (same path) | 20 | — | 2.957ms | — | 7.424ms | 2.120 ms |

## Why

Reduce latency and allow sharing robot over same host or over the network.

## Validation

-

## Breaking changes

- Introduces `device_ids` to the Robot protocol. All existing and future robot driver must implement it. 

## Related issues

- Closes #  https://github.com/openvinotoolkit/physicalai/issues/180


src/physicalai/robot/transport/_ids.py
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        robot_kwargs: Constructor kwargs containing ``port`` or ``ip``.
+
+    Returns:
+        A device id string suitable for key and lock-file naming.
+
+    Raises:
+        ValueError: If neither ``port`` nor ``ip`` is present.
+    """
+    port = robot_kwargs.get("port")
```
MarkRedeman: I think this might be a bit brittle and assumes too much of how the robot internally works.
For example a user could implement a `BimanualWidowXAI` robot as taking ` left_ip` and `right_ip` (versus the current implementation where a user is passing an already constrcuted `left = WidowXAI` and `right = WidowXAI` robot).

Perhaps we could explicitly provide an id? Or maybe we can add a new `Identifeable` protocol that users can implement giving an `id()`?

maxxgx: Yeah port/ip detection is not that reliable. Currently, this is a fallback for when `robot_id` is not supplied to `SharedRobot`.

I don't think we should require custom names/IDs so maybe we could add an `id` property to the robot protocol? We could create another protocol but it seems overly complicated for a single field/property.

What do you think of:
```python
@property
def device_id(self) -> str:
    """Stable identifier for this robot's physical hardware.

    MUST be deterministic and device-bound: the same physical robot must
    return the same value across processes and restarts (it keys the
    shared-transport topics and lock file). MUST NOT be random or
    per-process (no uuid4). Typically derived from the connection
    endpoint (resolved serial port, IP) or a device serial number.
    """
    ...
```

maxxgx: Added new Robot protocol property: `device_ids`, a tuple containing all the hardware ids that the robot uses

tests/unit/runtime/test_telemetry.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (C) 2025-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
+# ruff: noqa: ARG005, D100, D101, D102, E741, F401, I001, PLC2701, PLR2004, PLR6301, S101, SLF001, TC002, TC003
```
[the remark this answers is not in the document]
maxxgx: fine for test module

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+`name` that finds no existing owner spawns one (in a detached subprocess);
+later instances (same or different process, same `name`) attach to it:
+
+```python
+import numpy as np
+from physicalai.robot import SharedRobot
+from physicalai.robot.so101 import SO101
+
+robot = SharedRobot(
+    "left-arm",
```
MarkRedeman, line 28: General note that we discussed offline: for studio integration we will likely need to use `robot.id` as its name as we don't guarantee uniqueness of a robot name.

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+
+Notes:
+
+- `get_observation()` returns the newest owner-published state; if no new
+  sample arrived since the last call, the cached last-known observation is
+  returned. Staleness is visible via `obs.timestamp`.
+- `obs.state` is the **owner-computed** state vector (e.g. WidowXAI ships
+  positions + velocities, 14 values), so model inputs are correct without
+  robot-specific logic on the subscriber side.
+- `obs.images` is always `None` — camera frames go through `SharedCamera`
```
MarkRedeman, line 56: I feel like this might emphasize that it may be nice to simply remove cameras from the robot interface. At least for studio I haven't found a use case for it, and these type of restrictions could be unexpected to users.
If we remove them then there will be less surprises for users.

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+  robot.connect()
+
+  # Later discovery uses the already-established remote session.
+  robots = client.discover()
+```
+
+`SharedRobotClient` is attach-only: it never starts an owner process. It
+disconnects every robot created through `attach()` before closing its shared
+session on context exit. Its first `discover()` call uses a one-second budget
+for Zenoh scouting; later calls use 0.1 seconds with the warmed session.
```
MarkRedeman, line 105: I don't fully get this, why do we take 1 second for the first time?

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+with SharedRobotClient(allow_remote=True) as client:
+  robots = client.discover()
+  robot = client.attach(robots[0]["name"])
+  robot.connect()
+
+  # Later discovery uses the already-established remote session.
+  robots = client.discover()
+```
+
+`SharedRobotClient` is attach-only: it never starts an owner process. It
```
MarkRedeman, line 102: How much faster is it to connect to a robot via a shared client vs doing this without a client?

I'm wondering if it would be good to a add a shared client to Studio's DI / lifecycle that can then be reused. But I'm not sure if it's worth it if we can't use it to start processes?

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+    "left-arm",
+    robot_class=SO101,
+    robot_kwargs={"port": "/dev/ttyUSB0", "calibration": "calibration.json"},
+    allow_remote=True,
+)
+```
+
+The caller that spawns the owner fixes this scope for the owner's entire
+lifetime — a later attacher's `allow_remote` only configures its own
+session, it never widens or narrows an already-running owner's
```
MarkRedeman: What does it mean for a attacher to configure its own sessoin, if it can't widen or narrow (= nothing?) the running session?

docs/development/robot-zenoh-transport-design.md
```diff
@@ -0,0 +1,345 @@
...
+Attach-only construction is explicit:
+
+```python
+robot = SharedRobot.attach("left-arm")
+```
+
+Two identities serve different purposes:
+
+- `name` is the logical service identity used in Zenoh keys. It is required and must
+  be one non-empty segment containing only ASCII letters, digits, `_`, or `-`.
```
MarkRedeman: Would allowing forward slashes `/` be something we could allow here? That way we could defeine `serial:/dev/ttyUSB0` as a name.

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,178 @@
...
+`name` is what you choose and what keys the Zenoh topics — it never
+requires constructing a driver to resolve. Physical device identity
+(`Robot.device_ids`, e.g. `("serial:ttyUSB0",)`) is a separate concern the
+_owner_ uses to enforce host-local exclusivity: two different `name`s
+cannot claim the same physical device at the same time
+(`RobotDeviceAlreadyOwned`), and a race for the _same_ `name` with
+_different_ devices is rejected (`RobotNameConflict`) rather than silently
+picked by whichever process happened to start first.
+
+An existing owner's advertised `robot_class` is compared against yours only
```
MarkRedeman, line 152: What about the payload / kwargs, do we log any differences here if two processes try to start the same type of shared robot?
Or would that trigger a `RobotNameConflict`?

skills/runtime/physicalai-runtime-adding-a-robot-integration/references/robot-protocol.md
```diff
@@ -16,6 +16,17 @@ Class attribute:
 
 - `joint_names: list[str]` — order matches `joint_positions` and training action layout.
 
+Property:
+
+- `device_ids: tuple[str, ...]` — canonical, scheme-qualified identities of every
+  physical device this instance exclusively owns (e.g. `("serial:ttyUSB0",)`,
+  `("tcp:192.168.1.2",)`). Must be derivable from constructor parameters alone,
```
MarkRedeman, line 23: In another comment I asked about allowing forward slashes in a name, I think that would be useful in this example because there may be an edge case where users add virtual usb ports that may not be in `/dev`, so you could have duplicated `ttyUSB0` entries living in two separate directories.
I'm not sure though how likely those edge cases would be. Doing a trade off here of keeping the simplicity of the current name format might also be worth it..?

skills/runtime/physicalai-runtime-adding-a-robot-integration/references/robot-protocol.md
```diff
@@ -16,6 +16,17 @@ Class attribute:
 
 - `joint_names: list[str]` — order matches `joint_positions` and training action layout.
 
+Property:
+
+- `device_ids: tuple[str, ...]` — canonical, scheme-qualified identities of every
```
MarkRedeman, line 21: Why is it `device_ids` and not `device_id`? (single `str`)

maxxgx: It accounts for robots that use multiple physical devices, e.g. bimanual SO101 uses 2 serial ports

src/physicalai/robot/transport/_importing.py
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Dotted-path imports for trusted robot owner configuration."""
+
+from __future__ import annotations
+
+import importlib
+
+
+def import_dotted_path(path: str) -> object:
```
MarkRedeman, line 11: Is there any difference to this vs the `import_dotted_path` from the inference module?

src/physicalai/robot/so101/so101.py
```diff
@@ -207,6 +209,12 @@ def joint_names(self) -> list[str]:
         """List of joint names in the order they appear in state/action arrays."""
         return self.JOINT_ORDER
 
+    @property
+    def device_ids(self) -> tuple[str, ...]:
+        """Symlink-resolved serial device id, e.g. ``("serial:ttyUSB0",)``."""
+        name = Path(self._port).resolve().name if self._port.startswith("/dev/") else Path(self._port).name
+        return (f"serial:{name}",)
```
MarkRedeman: Might be nice to extract this to a `device_id_from_serial_port()` method. Then plugins can import this as a helper.

src/physicalai/robot/transport/_client.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def __init__(self, *, allow_remote: bool = False) -> None:
+        """Create a remote shared-robot client without opening a session yet.
+
+        Raises:
+            ValueError: If remote transport was not explicitly enabled.
+        """
+        if not allow_remote:
+            msg = "SharedRobotClient requires allow_remote=True"
+            raise ValueError(msg)
```
MarkRedeman: Initially I assumed that the client would also be usable to find local processes.
Am not sure if that would make sense though, but might want to clarify that in the docs?

docs/how-to/runtime/share-a-robot.md
maxxgx: The name takes priority, so if there's already a robot with name "r1", then we attach to it and log a warning to let the user know. 

`RobotNameConflict` is triggered only when there is a concurrent event where 2 robots with the same name but different device_ids are created at the same time. It should be a rare event.

src/physicalai/robot/transport/_client.py
```diff
@@ -0,0 +1,130 @@
...
+        self._session: Any = None
+        self._robots: list[SharedRobot] = []
```
MarkRedeman, line 37: Perhaps call this `self._attached_robots` to emphasize that these are attached robots vs discovered robots?

src/physicalai/robot/transport/_client.py
```diff
@@ -0,0 +1,130 @@
...
+
+from ._session import open_session
+from ._shared_robot import SharedRobot, discover_robots
+
+_COLD_DISCOVERY_TIMEOUT = 1.0
+_WARM_DISCOVERY_TIMEOUT = 0.1
+_RETRY_INTERVAL = 0.2
+
+
+class SharedRobotClient:
```
MarkRedeman, line 21: Should this also include a method to close a connection to a specific robot?
What happens when the consumer disconnects from the robot?

I.e.
```python
robot = client.attach(..)

# Later..
robot.disconnect()

# client is unaware of the disconnect
```

Are there use cases where it would be helpful for the client to be aware of this? (since it does keep a reference in `self._robots`)

src/physicalai/robot/transport/_session.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    import zenoh  # noqa: PLC0415
+
+    config = zenoh.Config()
+    config.insert_json5("mode", '"peer"')
+    if not allow_remote:
+        config.insert_json5("scouting/multicast/enabled", "false")
+        config.insert_json5("scouting/gossip/enabled", "false")
+    if name is not None:
+        port = derive_endpoint_port(name)
```
MarkRedeman, line 59: There's a very minimal chance that the port used here could collide with another name. I'm wondering if perhaps it's useful to include an optional `port` argument so that users may overwrite the port used.
Besides avoiding/resolving conflicts that could allow people be very specific about what port ranges they may want to use?

I'm also ok with keeping this here, but wondering if it would be good to document this implicit behaviour.

docs/development/robot-zenoh-transport-design.md
maxxgx: Technically I believe it's possible, but I won't recommend that as our zenoh keys contains `/` already, e.g. `physicalai/robot/left-arm/state`

src/physicalai/robot/interface.py
```diff
@@ -137,3 +137,21 @@ def joint_names(self) -> list[str]:
...
+    @property
+    def device_ids(self) -> tuple[str, ...]:
+        """Canonical identities of every physical device this instance exclusively owns.
+
+        Available immediately after construction, before :meth:`connect` is
+        called — must be derivable from constructor parameters alone, with
+        no hardware I/O. Stable across processes, restarts, and equivalent
+        connection aliases (e.g. a symlinked serial path).
+
+        Single-device robots return one id; composite robots (e.g. a
```
MarkRedeman, line 150: For composite robots would returning a join of each of its internal robot ids not work here vs a tuple?

I.e. `{self.left.device_id}-{self.right.deviceid}`.

Asking because on first read it felt a bit weird that the interface is called `Robot` (singular) but this method implies plural.

docs/how-to/runtime/share-a-robot.md
maxxgx: Probably could've explained better. It just means that every subscriber creates its own [zenoh session](https://docs.rs/zenoh/latest/zenoh/struct.Session.html)

src/physicalai/runtime/observer/_codec.py
```diff
@@ -0,0 +1,70 @@
...
+"""Msgpack and NumPy encoding for runtime observer telemetry."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import numpy as np
+
+
+def encode_numpy(array: np.ndarray) -> dict[str, Any]:
```
MarkRedeman, line 13: The codec logic here is duplicated from the robot's transport codec logic.
Is that intentional? Or could we extract this to a `codec` module that can get reused?

docs/how-to/runtime/share-a-robot.md
maxxgx: updated this part's wording

docs/how-to/runtime/share-a-robot.md
maxxgx: I've added this as option thing because I observed that initializing a zenoh session takes some time before the actual remote connections can be established. In my experiment, it took 0.8s for a remote reply to come in from the moment the session is started. However, using a "warm" session it only takes <0.1s to discover robots on the network. Within the same host, this is not an issue.

docs/how-to/runtime/share-a-robot.md
maxxgx: See, below. In short, a "cold" start (fresh zenoh session) takes some time to find all the connections

docs/how-to/runtime/share-a-robot.md
maxxgx: I thought about it and it seems a good idea. Then we also remove the type dependency on `Frame` from the capture submodule

src/physicalai/robot/transport/_importing.py
maxxgx: No, but I didn't want to share the code as it's better to keep the subpackages independent from each other. I think a little code duplication is okay here

src/physicalai/runtime/observer/_codec.py
maxxgx: Yeah, same idea as `import_dotted_path`.

src/physicalai/robot/transport/_lock.py
```diff
@@ -0,0 +1,280 @@
...
+        return None
+    if not isinstance(diagnostics, dict) or diagnostics.get("kind") != NAME_KIND:
+        return None
+
+    name = diagnostics.get("identity")
+    pid = diagnostics.get("pid")
+    if not isinstance(name, str) or not isinstance(pid, int) or pid <= 0:
+        return None
+    try:
+        os.kill(pid, 0)
```
MarkRedeman, line 107: The docblock isn't implying that `os.kill` will get called, why do we do this?

Edit: Oh I guess `kill -0` only checks that the process exists? Maybe add a small comment here for clarification.

src/physicalai/robot/transport/_owner.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        proc = self._process
+        if proc is None or proc.poll() is not None:
+            self._process = None
+            return
+
+        proc.terminate()
+        try:
+            proc.wait(timeout=timeout)
+        except subprocess.TimeoutExpired:
```
MarkRedeman, line 139: Would it be useful to add a `logger.warning` here?

src/physicalai/robot/transport/_owner_config.py
```diff
@@ -0,0 +1,163 @@
...
+        rate_hz: Owner loop rate.
+        idle_timeout: Seconds with zero subscribers before self-exit.
+    """
+
+    name: str
+    robot_class: str
+    robot_kwargs: dict[str, Any] = field(default_factory=dict)
+    allow_remote: bool = False
+    rate_hz: float = DEFAULT_RATE_HZ
+    idle_timeout: float = 10.0
```
MarkRedeman, line 98: For studio I'd like to do some experiments to see how fast we can spin up these processes. If it takes some time then we could increase the timeout a bit more, perhaps 1 or 2 minutes so that when the user switches between pages that they can then get an immediate reconnect.

src/physicalai/robot/transport/_client.py
maxxgx: I think it should be fine as the client only keeps the reference for lifecycle management (calls disconnect on exit) and disconnect is idempotent.

src/physicalai/robot/transport/_client.py
maxxgx: Yeah makes sense to allow it for local as well. Updated. Docs still focus on remote since it's the main benefit of using the client

src/physicalai/robot/interface.py
maxxgx: I think a single id is more fragile as we have a lock per device. I don't find a robot that is composed of multiple devices to be weird

MarkRedeman (dismissed):

src/physicalai/robot/so101/so101.py
maxxgx: Created the helper function under `device_ids.py`

src/physicalai/robot/transport/_codec.py
```diff
@@ -0,0 +1,218 @@
...
+def _encode_numpy(array: np.ndarray) -> dict[str, Any]:
+    return {
+        "__np__": True,
+        "dtype": str(array.dtype),
+        "shape": list(array.shape),
+        "data": array.tobytes(),
+    }
+
+
+def _decode_payload(value: object) -> object:
```
AlexanderBarabanov, line 52: Should we add some input validation to prevent potential logic errors/denial of service, e.g.:
- cap `len(data)` with some max value
- validate `len(data)` is consistent with `dtype.itemsize * product(shape)` before allocation.
- add a maximum nesting depth guard in `_decode_payload`.

src/physicalai/robot/transport/_owner_worker.py
```diff
@@ -0,0 +1,469 @@
...
+    period = 1.0 / rate_hz
+    idle_since: float | None = None
+    consecutive_failures = 0
+    next_tick = time.monotonic()
+
+    while not shutdown.is_set():
+        sample = action_sub.try_recv()
+        if sample is not None:
+            try:
+                action, goal_time, _send_ts = decode_action(sample.payload.to_bytes())
```
AlexanderBarabanov, line 202: Should we add some validation for returned  numpy array before use it in `send_action` (e.g, `dtype`, shape, values, probably)? 
Or is it out of this PR scope?

src/physicalai/robot/transport/_shared_robot.py
```diff
@@ -0,0 +1,559 @@
...
+
+    Returns:
+        One decoded metadata dict per answering owner, each including its
+        ``name``.
+    """
+    robots: list[dict[str, Any]] = []
+    deadline = time.monotonic() + timeout
+
+    def _append_replies(query_session: Any, query_timeout: float) -> None:  # noqa: ANN401
+        replies = query_session.get(METADATA_WILDCARD, timeout=query_timeout)
```
AlexanderBarabanov, line 124: If technically possible, please consider a configurable metadata disclosure level (e.g., omit `device_ids` and `host` from `/metadata` when `allow_remote=True`).

Also, please document in the user-facing how-to guide (docs/how-to/runtime/share-a-robot.md) that `discover_robots(allow_remote=True)` exposes inventory information.

src/physicalai/robot/transport/_owner_worker.py
```diff
@@ -0,0 +1,469 @@
...
+def signal_error(
+    msg: str,
+    tb: str | None = None,
+    *,
+    phase: str | None = None,
+    device_ids: tuple[str, ...] | None = None,
+) -> None:
+    payload: dict[str, Any] = {"msg": msg}
+    if tb:
+        payload["traceback"] = tb
```
AlexanderBarabanov, line 81: Minor: if the parent process logs this error to a centralized logging system or if the process stdout is captured in a multi-user/tenant environment, internal implementation details may leak.

Can we truncating/sanitizing tracebacks, at minimum ensure that `robot_kwargs` values (which may contain file paths or IP addresses) are not captured in tracebacks?

docs/how-to/runtime/share-a-robot.md
```diff
@@ -0,0 +1,180 @@
...
+  honoring the safe-state contract (hold/home) — and exits.
+- A subscriber's `disconnect()` never stops the robot's motors; the owner
+  owns safe-state.
+- Subscribers reject an owner advertising an unsupported transport
+  protocol version before ever declaring the action publisher.
+
+## Security: trusted network required
+
+In `allow_remote=True` mode, this transport applies any action received on
+its Zenoh `/action` key **without authentication** — any peer that can
```
AlexanderBarabanov:
```suggestion
its Zenoh `/action` key **without authentication and encryption** — any peer that can
```

src/physicalai/robot/transport/_codec.py
maxxgx: Added `MAX_PAYLOAD_BYTES=1MiB` check and wrapped `decode_state` in try-except  . `msgpack` already enforces max recursion depth, so I think we don't need to add the same mechanism here

src/physicalai/robot/transport/_owner_worker.py
maxxgx: I don't think it's needed as shape and value verification is already the driver's responsibility

src/physicalai/robot/transport/_shared_robot.py
maxxgx: I've omitted `device_ids` for remote robots, but I think it's quite useful to keep `host` for debugging and to discover robots controlled by a specific machine.

src/physicalai/robot/transport/_owner.py
maxxgx: added

src/physicalai/robot/transport/_lock.py
maxxgx: added

skills/runtime/physicalai-runtime-adding-a-robot-integration/references/robot-protocol.md
maxxgx: I would leave it as is for now and maybe re-evaluate it when we hit a real edge case

src/physicalai/robot/transport/_owner_worker.py
maxxgx: I think on this should be already isolated in the parent process as we only write to `stdout`

AlexanderBarabanov (approved):

PR #196, opened 2026-07-16
chore(deps): bump `transformers` >=5.6.0,<5.7.0

AlexanderBarabanov:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "pyyaml>=6.0",
     "opencv-python-headless",
     "huggingface_hub>=0.34,<2.0",
-    "transformers>=5.5.0,<5.6.0",
+    "transformers>=5.6.0,<5.7.0",
     "num2words>=0.5.14,<0.6.0",
     "safetensors>=0.4.3,<1.0.0",
     "onnxruntime",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "pyyaml>=6.0",
     "opencv-python-headless",
     "huggingface_hub>=0.34,<2.0",
-    "transformers>=5.5.0,<5.6.0",
+    "transformers>=5.6.0,<5.7.0",
```
sovrasov, line 30: We have to stay compatible with lerobot and can not really move the lower bound at this point:
https://github.com/huggingface/lerobot/blob/main/pyproject.toml#L152

maxxgx: We also need to consider what version PhysicalAI Studio library needs as it currently uses `"transformers>=5.3.0,<5.4.0"`. This already makes latest studio and framework incompatible with each other

AlexanderBarabanov: Thanks for point to lerobot. 
For physical AI studio - this will be done in https://github.com/open-edge-platform/physical-ai-studio/pull/808/

AlexanderBarabanov:
Closing due to https://github.com/huggingface/lerobot/blob/main/pyproject.toml#L152 restriction.

PR #197, opened 2026-07-16
fix: Improve 004 robotless deployment tutorial with dataset and model from official repo and MuJoCo visualization

zhuo-yoyowz:
## Summary

-This PR improves 004_Test_Deployment_Without_Robot.ipynb so users can run the full robotless deployment flow more smoothly from a fresh environment, with dataset and Pi0.5 model in IR format from official repo, and observe final result in MuJoCo env.

## Changes

-Add first-run setup logic so the notebook can clone the Physical AI tutorial sources when required files are missing.
Install physicalai==0.1.1 and required runtime dependencies for a more stable tutorial experience.
-Add LIBERO dataset replay preparation using HuggingFaceVLA/libero, matching the released OpenVINO/pi05-libero-fp16-ov model.
-Add replay validation for downloaded dataset frames, task text, expert actions, and model-predicted actions.
-Keep OpenVINO device selection for available devices, preferring GPU when visible.
-Add a MuJoCo-based scripted pick-and-place visualization as a final robotless smoke test.
-Add helper utilities in physicalai_pi05_helper.py for dataset download, replay visualization, model execution, and MuJoCo rendering.


examples/tutorials/physicalai_pi05_helper.py
maxxgx, line 1: I don't think we should keep this source file. I guess the 2D visualization in 001 used this to generate the GIF with the "simulated" robot arm. But that didn't make any sense as it was an arbitrary representation (even with fewer joints) and not based on the physical properties of the robot like an URDF.

maxxgx (commented):
I'm not sure about the Mujoco part.

Libero runs a Mujoco environment under the hood, if we can use that simulation it would be great and we could avoid an ad-hoc Mujoco section. Could you please explore if we can get that from Libero?

In PAI Studio we have a `VideoRecorder` under the library and `LiberoBenchmark` takes a `video_dir` and `record_mode`. Maybe we can just run the benchmark and show a video afterwards?

zhuo-yoyowz:
This update modernizes notebook 004 and addresses the latest review feedback:

- Replaced the custom MuJoCo helper with the official LiberoBenchmark API from the Physical AI Studio library.
- Uses the latest PhysicalAI inference API to load Hugging Face or local OpenVINO policy packages.
- Runs genuine closed-loop policy evaluation in the LIBERO MuJoCo environment.
- Records and displays rollout videos generated from OpenVINO inference actions.
- Added clear explanations of how OpenVINO inference, LIBERO, MuJoCo, and video recording work together.
- Added configurable task, episode, seed, recording, and OpenVINO device settings.
- Added Linux prerequisite guidance, including FFmpeg, EGL, compiler, and CMake requirements.
- Added a CMake compatibility workaround for egl_probe.
- Removed the obsolete physicalai_pi05_helper.py.
- Cleared notebook outputs and execution counts before submission.
- Validation completed on Ubuntu 24.04, including dependency installation, model loading, OpenVINO GPU inference, LIBERO benchmark execution, MuJoCo rendering, and MP4 playback.

zhuo-yoyowz:
Addressed the latest validation and environment setup issues for notebook 004.
### What changed
- Moved the heavy LIBERO and Physical AI Studio installation out of the notebook
 and into an explicitly optional, notebook-004-only section in the tutorials
 README.
- Kept the shared `uv sync --frozen --extra notebooks` environment lightweight
 for tutorials that do not require LIBERO.
- Added the required CMake 3.31.10 and `hf-egl-probe` build setup for Ubuntu 24.04.
- Installed Physical AI Studio and the latest PhysicalAI runtime sequentially to
 avoid uv's conflicting Git URL resolution.
- Added a post-install LIBERO configuration step so its resource paths point to
 the package installed in the current `.venv`, rather than a stale virtual
 environment.
- Clarified creation and selection of the dedicated
 `PhysicalAI Tutorials (uv)` Jupyter kernel.
- Simplified the first notebook cell so it only verifies dependencies and the
 required `InferenceModel.from_pretrained()` API. It no longer modifies the
 environment while the kernel is running.
- Added guidance for using `https://hf-mirror.com` when Hugging Face access from
 the PRC fails because of connection issues.
- Restored the dedicated uv kernelspec and cleared notebook outputs and execution
 counts.
### Why
Installing the simulation stack inside the first notebook cell could take a long
time, produce excessive Jupyter output, leave the running kernel with stale
namespace-package imports, and make LIBERO retain paths from an older virtual
environment. Preparing these dependencies before Jupyter starts makes the uv
environment deterministic and keeps notebook execution focused on the workflow.
The LIBERO dependencies remain optional and scoped specifically to notebook 004,
so future tutorials do not need to install the simulation stack unless required.
### Validation
- Completed the full notebook workflow successfully on Ubuntu 24.04 using the
 repository `.venv` and the dedicated uv kernel.
- Verified model loading, OpenVINO device selection, LIBERO benchmark execution,
 and MuJoCo rollout recording.
- Verified notebook JSON structure, Python cell syntax, clean outputs,
 kernelspec metadata, and `git diff --check`.

maxxgx (approved):

PR #202, opened 2026-07-27
feat(config): add core component config engine

maxxgx:
## Summary
- Introduces `physicalai.config` (`@export_config`, `to_config`, `instantiate`, normalization, envelope validation, cycle detection).
- Consolidates dotted-path importing under `physicalai.config.importing`; inference and robot transport re-export from there.
- Updates `docs/reference/config-schema.md` and adds unit tests under `tests/unit/config/`.

## Test plan
- [x] `uv run pytest tests/unit/config/`

### Review & merge order

1. [PR #202 — Core config engine (this PR)](https://github.com/openvinotoolkit/physicalai/pull/202) → `main`  
   Core `physicalai.config` (export, instantiate, validation)

2. [PR #204 — Robot drivers & SharedRobot transport](https://github.com/openvinotoolkit/physicalai/pull/204)  
   Robot config export and SharedRobot owner process startup

3. [PR #205 — Camera drivers & SharedCamera transport](https://github.com/openvinotoolkit/physicalai/pull/205)  
   Camera config export and SharedCamera publisher startup

4. [PR #206 — Runtime, InferenceModel & CLI](https://github.com/openvinotoolkit/physicalai/pull/206)  
   End-to-end YAML, `physicalai run --config`, examples, and user docs

**In parallel:** [PR #203 — Component config design doc](https://github.com/openvinotoolkit/physicalai/pull/203) → `docs/design`  
Detailed design doc; not blocking the code PRs

### DIFF
--- a/src/physicalai/config/_instantiate.py
+++ b/src/physicalai/config/_instantiate.py
@@ -1,0 +1,276 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Instantiate trusted ComponentConfig trees."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Mapping
+from enum import Enum
+from typing import cast
+
+from ._errors import ComponentConfigError, ComponentImportError
+from ._export import declared_config_args
+from ._normalize import validate_component_config
+from ._path import format_path
+from ._types import _MAX_CONFIG_DEPTH, ComponentConfig, JsonValue
+from .importing import import_dotted_path
+
+
+def _is_nested_config(value: object) -> bool:
+    return isinstance(value, Mapping) and "class_path" in value
+
+
+def _check_preflight_depth(path: str, depth: int) -> None:
+    if depth > _MAX_CONFIG_DEPTH:
+        msg = f"{format_path(path)}: nesting depth exceeds {_MAX_CONFIG_DEPTH}"
+        raise ComponentConfigError(msg)
+
+
+def _preflight_config(
+    config: object,
+    *,
+    path: str,
+    depth: int,
+    seen: set[int],
+) -> None:
+    """Validate a complete component subtree before any class import.
+
+    Raises:
+        ComponentConfigError: If the config tree is malformed.
+    """
+    _check_preflight_depth(path, depth)
+
+    validated = validate_component_config(config, path=path)
+    config_id = id(config)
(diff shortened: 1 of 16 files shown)


src/physicalai/config/_types.py
```diff
@@ -0,0 +1,45 @@
...
+    """Domain value that encodes to constructor-compatible JSON for export."""
+
+    def to_config_value(self) -> JsonValue:
+        """Return ctor-compatible JSON for component-config export."""
+        ...
+
+
+# Maximum nesting depth for recursive normalization and instantiation.
+# Counts traversal through lists and mappings as well as nested component configs.
+_MAX_CONFIG_DEPTH = 10
```
MarkRedeman, line 32: Perhaps make this configurable via an environment variable with default 10?

src/physicalai/config/_instantiate.py
```diff
@@ -0,0 +1,276 @@
...
+    raise ComponentConfigError(msg)
+
+
+def _decode_value(
+    value: object,
+    *,
+    path: str,
+    depth: int,
+    seen: set[int],
+) -> object:
```
MarkRedeman, line 151: In here there are quite a few:
```
if (conditoin)
   length body
   return thing
```

perhaps that could be simplified by using some polymorphism / strategy pattern here?
Though likely that this would also make things more complicated..

MarkRedeman (approved):
Some very minor suggestions that we can probably ignore.. I tried going over all of the code but did find it difficult to read some of it, it feels very llm'y.
I'm not sure what improvements to suggest though, conceptually it all looks really good, but its sometimes hard to track some of the assumptions being made in the code.

One thing I was thinking of is that sometimes we have code like,
```python

def normalize_value(
    value: object,
    *,
    path: str = "",
    depth: int = 0,
    seen: set[int] | None = None,
    codec_seen: set[int] | None = None,
    to_config: ToConfigFn | None = None,
    is_exportable: IsExportableFn | None = None,
    domain_codec: DomainCodecFn | None = None,
) -> JsonValue:
```

the `seen` and `codec_seen` feel like properties specific to this normalization. Perhaps placing them together into one dataclass, or having a `Normalizer` class that holds these values internally, could help clean up things.

src/physicalai/config/_types.py
maxxgx: Will consider it in one of the followup PRs

PR #204, opened 2026-07-27
feat(robot): export config and cut over SharedRobot transport

maxxgx:
## Summary
- Decorates SO101 and WidowX drivers with `@export_config`.
- Cuts over robot owner stdin startup envelopes to `robot: ComponentConfig`.
- Updates `SharedRobot.from_config()`, owner worker, and `physicalai robot serve` CLI.

Stacked on #202.

## Test plan
- [x] `uv run pytest tests/unit/robot/test_robot_component_config.py tests/unit/robot/transport/ tests/unit/cli/test_robot_cli.py`

### Review & merge order

1. [PR #202 — Core config engine](https://github.com/openvinotoolkit/physicalai/pull/202) → `main` [MERGED] 
   Core `physicalai.config` (export, instantiate, validation)

2. [PR #204 — Robot drivers & SharedRobot transport (this PR)](https://github.com/openvinotoolkit/physicalai/pull/204)  
   Robot config export and SharedRobot owner process startup

3. [PR #205 — Camera drivers & SharedCamera transport](https://github.com/openvinotoolkit/physicalai/pull/205)  
   Camera config export and SharedCamera publisher startup

4. [PR #206 — Runtime, InferenceModel & CLI](https://github.com/openvinotoolkit/physicalai/pull/206)  
   End-to-end YAML, `physicalai run --config`, examples, and user docs

**In parallel:** [PR #203 — Component config design doc](https://github.com/openvinotoolkit/physicalai/pull/203) → `docs/design`  
Detailed design doc; not blocking the code PRs

### DIFF
--- a/src/physicalai/cli/robot.py
+++ b/src/physicalai/cli/robot.py
@@ -62,8 +62,8 @@ def _build_serve_parser() -> ArgumentParser:
         required=True,
         help=(
             "Robot ComponentConfig (class_path + init_args). "
-            "Example: --robot='{\"class_path\":\"physicalai.robot.SO101\","
-            "\"init_args\":{\"port\":\"/dev/ttyACM0\"}}'. "
+            'Example: --robot=\'{"class_path":"physicalai.robot.SO101",'
+            '"init_args":{"port":"/dev/ttyACM0"}}\'. '
             "Or nested flags: --robot.class_path=physicalai.robot.SO101 "
             "--robot.init_args.port=/dev/ttyACM0"
         ),
--- a/src/physicalai/cli/robot.py
+++ b/src/physicalai/cli/robot.py
@@ -17,6 +17,7 @@ from jsonargparse import ActionConfigFile, ArgumentParser
 from loguru import logger
 
 from physicalai.cli._spec import SubcommandSpec  # noqa: PLC2701
+from physicalai.config import ComponentConfigError
 from physicalai.robot.errors import RobotTransportError
 from physicalai.robot.transport import (
     DEFAULT_RATE_HZ,
@@ -55,12 +56,11 @@ def _build_serve_parser() -> ArgumentParser:
     parser = ArgumentParser(description=_SERVE_HELP)
     parser.add_argument("--config", action=ActionConfigFile, help="YAML/JSON config file.")
     parser.add_argument("--name", type=str, required=True, help="Robot logical name.")
-    parser.add_argument("--robot_class", type=str, required=True, help="Trusted dotted path to the driver class.")
     parser.add_argument(
-        "--robot_kwargs",
+        "--robot",
         type=dict,
-        default=None,
-        help="JSON-serializable driver constructor arguments.",
+        required=True,
+        help="Trusted robot ComponentConfig (class_path + init_args).",
     )
     parser.add_argument(
         "--allow_remote",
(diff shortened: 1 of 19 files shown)


src/physicalai/cli/robot.py
```diff
@@ -55,12 +56,11 @@ def _build_serve_parser() -> ArgumentParser:
...
     parser.add_argument("--name", type=str, required=True, help="Robot logical name.")
-    parser.add_argument("--robot_class", type=str, required=True, help="Trusted dotted path to the driver class.")
     parser.add_argument(
-        "--robot_kwargs",
+        "--robot",
         type=dict,
-        default=None,
-        help="JSON-serializable driver constructor arguments.",
+        required=True,
+        help="Trusted robot ComponentConfig (class_path + init_args).",
```
MarkRedeman: Can we add an example here? I imagine this would require a json string?

src/physicalai/robot/transport/_owner_config.py
```diff
@@ -9,12 +9,18 @@
...
-Security: *robot_class* is trusted local application/config input, exactly
+handle cannot cross a process boundary (D15). Only a trusted
+:class:`~physicalai.config.ComponentConfig` (``class_path`` + ``init_args``)
+survives that boundary — arbitrary robot types, including third-party
+plugins, work without any registry lookup here.
+
+Private stdin is ``robot: ComponentConfig`` only. The owner envelope is
+validated schema-positively: required ``robot``, known transport keys, and
+rejection of unknown keys (including legacy flat ``robot_class`` /
+``robot_kwargs``) before import or hardware access. Public ``SharedRobot``
```
MarkRedeman: Do we need to keep handing a "legacy flat robot_class" etc?

src/physicalai/robot/transport/_shared_robot.py
```diff
@@ -177,17 +187,25 @@ class SharedRobot:
...
             instances constructed with the same *name* (anywhere reachable
             under the chosen transport scope) share one owner.
-        robot_class: Driver class (or its dotted import path) to spawn if
-            no owner exists yet for *name*. ``None`` means attach-only —
-            use :meth:`attach` for that case instead of passing this
-            directly.
-        robot_kwargs: JSON-serializable driver constructor kwargs (e.g.
-            ``port``, ``calibration`` as a path, ``role`` as a str), used
-            only when this instance spawns the owner.
+        robot: Trusted driver :class:`~physicalai.config.ComponentConfig` to
```
MarkRedeman: It's not entirely clear to me what a "trusted" driver would be

src/physicalai/robot/transport/_shared_robot.py
```diff
@@ -253,7 +312,15 @@ def attach(
         Returns:
             A ``SharedRobot`` that never spawns an owner.
         """
-        return cls(name, allow_remote=allow_remote, connect_timeout=connect_timeout, _session=_session)
+        # Omit default None so @export_config does not capture "_session": null.
+        if _session is not None:
+            return cls(name, allow_remote=allow_remote, connect_timeout=connect_timeout, _session=_session)
+        return cls(name, allow_remote=allow_remote, connect_timeout=connect_timeout)
```
MarkRedeman: The constructor has `_session: object | None = None`, so wouldn't ` cls(name, allow_remote=allow_remote, connect_timeout=connect_timeout)` be the same as ` cls(name, allow_remote=allow_remote, connect_timeout=connect_timeout, _session=_session)`, if `_session` is none?

tests/unit/robot/transport/fake.py
```diff
@@ -35,6 +37,7 @@ def state(self) -> np.ndarray:
         return self.joint_positions
 
 
+@export_config
```
MarkRedeman, line 40: Why no class path arg here?

MarkRedeman (commented):
This looks good to me, though I think some of the code could be simplified if we ignore the "legacy flat robot class"?

tests/unit/robot/transport/fake.py
maxxgx: without the `class_path` it defaults to module + class_name, which is okay for the test. So `class_path=` is optional, but it's good to have for stable public classes

src/physicalai/robot/transport/_shared_robot.py
maxxgx: At runtime it works without the branching. However, we re-export the sharedrobot via `@export_config`, then we will have `_session: null` in the config. So the branching fixes that

src/physicalai/robot/transport/_shared_robot.py
maxxgx: True, not very clear. It basically means that it doesn't come from the network. I've changed it "local" instead

src/physicalai/robot/transport/_owner_config.py
maxxgx: I updated the `SharedRobot` constructor, but yeah there is no legacy because we haven't release yet. Updated docstring.

src/physicalai/cli/robot.py
maxxgx: Correct, added example.
We accept a JSON string or nested flags like jsonargparse

MarkRedeman (approved):

PR #207, opened 2026-07-27
feat(inference): add ov_smoke integration test suite

pwolnows:
## Summary

Implements the `ov_smoke` pytest suite described in `docs/design/openvino-validation.md`. The test validates the three-stage contract _"OpenVINO works with physicalai"_ against the three published HuggingFace exports in the [OpenVINO/physical-ai](https://huggingface.co/collections/OpenVINO/physical-ai) collection.

## Motivation

An OpenVINO upgrade silently broke pi05 inference (only caught because of the existing `openvino==2026.1` pin). The root cause was an `openvino` / `openvino_tokenizers` version mismatch — ACT (no tokenizer) passed while pi05 failed at the tokenizer stage. Without an automated test this failure was invisible until runtime.

## Contract stages

| Stage | Code path | What it proves |
|---|---|---|
| **load** | `OpenVINOAdapter.load` | `read_model` + `compile_model` succeed |
| **tokenizer** | `OVTokenizer` | `openvino_tokenizers` extension loads and matches core OpenVINO |
| **predict** | `InferenceModel.predict_action_chunk` | inference runs, outputs are finite |

## Models

| Repo | Stages |
|---|---|
| `OpenVINO/act-fp16-ov` | load, predict (no tokenizer) |
| `OpenVINO/pi05-libero-fp16-ov` | load, **tokenizer**, predict |
| `OpenVINO/smolvla-libero-fp16-ov` | load, **tokenizer**, predict |

Verified: under a core/tokenizer mismatch `test_load[act-fp16-ov]` passes while `test_tokenizer[pi05-libero-fp16-ov]` fails — detecting the exact break that caused the pin.

## Changes

- `tests/integration/inference/test_ov_smoke.py` — 8-test suite (3 × load, 2 × tokenizer, 3 × predict)
- `pyproject.toml` — registers the `ov_smoke` pytest marker
- `docs/design/openvino-validation.md` — design proposal for the full OV validation pipeline (smoke, golden-action, perf, gym E2E)

## Usage

```bash
pytest -m ov_smoke -s
# with local model cache:
OV_SMOKE_CACHE_DIR=~/.cache/hf pytest -m ov_smoke
# target a single model:
pytest -m ov_smoke -k pi05
```


maxxgx (dismissed):

tests/integration/inference/test_ov_smoke.py
```diff
@@ -0,0 +1,318 @@
...
+    params=_ALL_MODELS,
+    ids=lambda m: m.short_id,
+)
+def all_exports(
+    request: pytest.FixtureRequest,
+    smoke_cache_dir: Path | None,
+) -> tuple[Path, _ModelSpec]:
+    """Download (or reuse cached) export dir for each model in *_ALL_MODELS*."""
+    spec: _ModelSpec = request.param
+    export_dir = download_from_hub(spec.repo_id, cache_dir=smoke_cache_dir)
```
maxxgx: Please pin the HF repo revision using `revision=`

tests/integration/inference/test_ov_smoke.py
```diff
@@ -0,0 +1,318 @@
...
+    params=_TOKENIZER_MODELS,
+    ids=lambda m: m.short_id,
+)
+def tokenizer_exports(
+    request: pytest.FixtureRequest,
+    smoke_cache_dir: Path | None,
+) -> tuple[Path, _ModelSpec]:
+    """Download (or reuse cached) export dir for tokenizer-bearing models."""
+    spec: _ModelSpec = request.param
+    export_dir = download_from_hub(spec.repo_id, cache_dir=smoke_cache_dir)
```
maxxgx: For the tokenizer, you can limit the download with `allow_patterns=["manifest.json", "tokenizer.xml", "tokenizer.bin"]` so that you won't need to download the full weights.

Please also add `revision` hash here as well

tests/integration/inference/test_ov_smoke.py
```diff
@@ -0,0 +1,318 @@
...
+from physicalai.inference.constants import TASK, TOKENIZED_PROMPT, TOKENIZED_PROMPT_MASK
+from physicalai.inference.manifest import Manifest
+from physicalai.inference.model import InferenceModel
+from physicalai.inference.preprocessors import OVTokenizer
+from physicalai.inference.utils._hub import download_from_hub
+
+log = logging.getLogger(__name__)
+
+# All tests in this module carry both markers.
+pytestmark = [pytest.mark.ov_smoke, pytest.mark.integration, pytest.mark.requires_download]
```
maxxgx, line 64: I think if we run `uv run pytest`, this smoke test will also run and download the models from HF. IMO, we should run them only when we mean it, so `uv run pytest -m ov_smoke`. 

I think you can exclude the smoke test by default in `pyproject.toml`:
```toml
[tool.pytest.ini_options]
addopts = [
  "--strict-markers",
  "--strict-config",
  "--showlocals",
  "-ra",
  "-m",
  "not ov_smoke",
]
```

tests/integration/inference/test_ov_smoke.py
```diff
@@ -0,0 +1,318 @@
...
+        observation={**_LIBERO_IMAGES, **_LIBERO_STATE, **_LIBERO_TASK},
+    ),
+    _ModelSpec(
+        repo_id="OpenVINO/smolvla-libero-fp16-ov",
+        has_tokenizer=True,
+        observation={**_LIBERO_IMAGES, **_LIBERO_STATE, **_LIBERO_TASK},
+    ),
+]
+
+_TOKENIZER_MODELS: list[_ModelSpec] = [m for m in _ALL_MODELS if m.has_tokenizer]
```
maxxgx, line 120: I think we should split default vs full coverage while keeping a single ov_smoke test suite:

 - Default (run on PAI repo, CPU, on every PR/OV-bump): run ACT load/predict + tokenizer stage only (download just manifest.json + tokenizer.xml/.bin for a tokenizer-bearing repo via `allow_patterns`).
- Full (run on OV repo, weekly/RC): run ACT + pi05 + smolvla across all stages with full model artifacts.

This would allow PAI to run the smoke test on a lightweight GHA runner (ACT + tokenizer only) without heavy downloads or compute on PAI, while still preserving full coverage in OV’s scheduled weekly runs.

maxxgx (changes requested):
Thanks for the PR!

I left a few comments, the main issue is that downloading full smolvla and Pi05 is too much as default (should be okay if run weekly on OV runner). 

It would also be nice to have the lightweight smoke test run on PAI and triggered by a change in `physical/inference` package, as well as on OV version bump (this can be followup PR).

maxxgx (approved):
Looks good, thanks!

PR #209, opened 2026-07-29
chore: CLI runtime logging

maxxgx:
## Summary

- Log model loading (otherwise it appears stuck)
- Add arg `--shared-robot` in example runtime scripts

## Why

-

## Validation

-

## Breaking changes

- None

## Related issues

- Closes #

### DIFF
--- a/examples/runtime/async_inference.py
+++ b/examples/runtime/async_inference.py
@@ -80,6 +80,15 @@ def main() -> None:
     robot_group.add_argument("--ip", help="Robot IP (widowxai)")
     robot_group.add_argument("--ip-left", help="Left arm IP (bimanual_widowxai)")
     robot_group.add_argument("--ip-right", help="Right arm IP (bimanual_widowxai)")
+    robot_group.add_argument(
+        "--shared-robot",
+        action="store_true",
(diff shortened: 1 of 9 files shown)


examples/runtime/async_inference.py
```diff
@@ -80,6 +80,15 @@ def _handle_sigint(sig: int, frame: object) -> None:
...
     robot_group.add_argument("--ip-left", help="Left arm IP (bimanual_widowxai)")
     robot_group.add_argument("--ip-right", help="Right arm IP (bimanual_widowxai)")
+    robot_group.add_argument(
+        "--shared-robot",
+        action="store_true",
+        help="Wrap the driver in SharedRobot (Zenoh transport)",
+    )
+    robot_group.add_argument(
+        "--robot-name",
+        help="Logical SharedRobot name (required with --shared-robot)",
```
MarkRedeman, line 90: Could the robot name not be inferred / given a default based on the `--robot` argument?

MarkRedeman (approved):

examples/runtime/async_inference.py
maxxgx: The script construct the robots differently than the config. It assumes either serial port or ip. Probably needs to be updated in future to support plugins

PR #210, opened 2026-07-29
feat(ci): add fuzzing

AlexanderBarabanov:
## Summary

This PR add fuzzing based on https://github.com/google/atheris
See `tests/fuzz/README.md` for the details.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,181 @@
+# Fuzzing for Physical AI Runtime
+
+name: Fuzz
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  workflow_dispatch:
+    inputs:
+      duration_s:
+        description: "Fuzzing duration per target (seconds)"
+        default: "300"
+        required: false
+  pull_request:
+    paths:
+      - "tests/fuzz/**"
+      - ".github/workflows/fuzz.yml"
+
+permissions: {}
+
+jobs:
+  fuzz:
+    name: "fuzz / ${{ matrix.harness }}"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Manifest / component instantiation
+          - harness: manifest
+            seed_dir: tests/fuzz/seeds/fuzz_manifest
+          - harness: component
+            seed_dir: ""
+          # Array processing pipeline
+          - harness: stats_normalizer
+            seed_dir: ""
+          - harness: resize_preprocessor
+            seed_dir: ""
+          - harness: action_normalizer
+            seed_dir: ""
+          - harness: action_chunk_trimmer
+            seed_dir: ""
+          - harness: lerp_smoother
+            seed_dir: ""
+          - harness: resize_smolvla
+            seed_dir: ""
+          # Path handling, runtime, queue
+          - harness: policy_name
+            seed_dir: ""
+          - harness: detect_backend
+            seed_dir: ""
+          - harness: prepare_inputs
+            seed_dir: ""
+          - harness: action_queue
+            seed_dir: ""
+          - harness: import_dotted_path
+            seed_dir: ""
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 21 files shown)


tests/fuzz/harnesses/fuzz_prepare_inputs.py
```diff
@@ -0,0 +1,160 @@
...
+    if dot_key in result_a and dot_key in result_b:
+        val_a = result_a[dot_key]
+        val_b = result_b[dot_key]
+        # At minimum the shapes must match — if the winner differs the shapes
+        # could differ, which is the actual safety bug (wrong tensor fed to model).
+        assert val_a.shape == val_b.shape, (
+            f"_prepare_inputs collision is non-deterministic: "
+            f"order-A produced shape {val_a.shape}, order-B produced {val_b.shape} "
+            f"for key {dot_key!r}"
+        )
```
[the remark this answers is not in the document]
AlexanderBarabanov: Works as expected and even detected some crash https://github.com/openvinotoolkit/physicalai/actions/runs/30530580889/job/90831700440?pr=210

tests/fuzz/harnesses/fuzz_component.py
```diff
@@ -0,0 +1,182 @@
...
+# Registry short names safe to instantiate for flat-params injection tests
+_SAFE_TYPES = [
+    "single_pass",
+    "action_chunk_trimmer",
+]
+
+
+@atheris.instrument_func
+def _sub_resolve_artifact(fdp: atheris.FuzzedDataProvider, export_dir: str) -> None:
+    """Assert that resolve_artifact never returns a path outside export_dir."""
```
maxxgx, line 35: I think `resolve_artifact` only works for relative paths. When a an absolute path is given, it returns unchanged. I guess that's why this test is failing. Is this an actual issue?

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,181 @@
...
+  workflow_dispatch:
+    inputs:
+      duration_s:
+        description: "Fuzzing duration per target (seconds)"
+        default: "300"
+        required: false
+  pull_request:
+    paths:
+      - "tests/fuzz/**"
+      - ".github/workflows/fuzz.yml"
```
maxxgx, line 17: Should changes in `src` also trigger the fuzz tests?

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,181 @@
...
+      # TODO: move atheris into pyproject.toml
+      - name: Install atheris
+        run: uv pip install "atheris==3.1.0"
+
+      - name: Seed corpus
+        run: |
+          mkdir -p "tests/fuzz/corpus/${{ matrix.harness }}"
+          if [ -n "${{ matrix.seed_dir }}" ] && [ -d "${{ matrix.seed_dir }}" ]; then
+            cp -rn "${{ matrix.seed_dir }}/." "tests/fuzz/corpus/${{ matrix.harness }}/" || true
+          fi
```
maxxgx, line 90: Do you save the corpus each after each run or does it always start over?

maxxgx (commented):
Thanks, Alexander. It's an interesting test framework and it seems it found some real issues already

.github/workflows/fuzz.yml
AlexanderBarabanov: Currently, it always start over, as an improvement we may implement some caching in future.

.github/workflows/fuzz.yml
AlexanderBarabanov: Currently the intention of running this on PRs is to catch cases where changes to fuzz tests or the fuzzing workflow could break existing fuzzing functionality. The fuzzing executed as part of a PR uses a short duration (passed via `-max_total_time`), unlike the scheduled nightly fuzzing runs.

I would propose running it only as a nightly job (and potentially increasing the duration in the future). We can then periodically analyze the findings and address any issues that are discovered.

tests/fuzz/harnesses/fuzz_component.py
AlexanderBarabanov: I think so, I will propose a fix in a separate PR.

maxxgx (approved):

PR #212, opened 2026-07-30
feat(config): unify Config API and absorb Studio config module

samet-akcay:
## Summary

- Consolidate the shared **`physicalai.config`** package in Runtime: typed dataclass `Config`, construction recipes (`@export_config`, strict `instantiate`), plus Studio-era helpers (`FromConfig`, `@from_config`, `instantiate_obj`, serialization).
- Finish the **`ComponentConfig` → `Config`** cutover across capture/robot transport, runtime, CLI, examples, and docs.
- Add config how-to docs (`instantiate-objects`, `use-from-config`), MkDocs nav, and agent skill `physicalai-runtime-working-with-config`.
- Expand and rename unit tests under `tests/unit/config/` (including FromConfig/instantiate coverage moved from Studio).

## Motivation

Physical AI Studio must consume config from the Runtime package only. This PR makes Runtime the single owner of implementation, documentation, tests, and agent skills for `physicalai.config`, unblocking a follow-up Studio PR that removes `library/src/physicalai/config/` and bumps the `physicalai` dependency.

## Follow-ups

- **Physical AI Studio**: merge after this lands; bump `[tool.uv.sources]` / `uv.lock` to this commit and remove any remaining Studio-only config scaffolding.

### DIFF
--- a/docs/explanation/configuration.md
+++ b/docs/explanation/configuration.md
@@ -5,7 +5,7 @@ can describe the same runtime graph.
 
 ## Construction Recipes
 
-A `ComponentConfig` names one class and the arguments needed to create it.
+A `Config` names one class and the arguments needed to create it.
 
 ```yaml
 class_path: physicalai.capture.UVCCamera
@@ -15,16 +15,17 @@ init_args:
   height: 480
 ```
 
-Classes opt in to exporting this recipe with `@export_config`. `to_config()`
+Classes opt in to exporting this recipe with `@export_config`.
+`Config.from_instance()`
 captures construction intent rather than mutable runtime state: supplied
 arguments are preserved, omitted defaults remain omitted, and connections or
 open handles are never exported.
 
 ```text
-live opted-in component -> to_config() -> JSON/YAML -> instantiate() -> new component
+live opted-in component -> Config.from_instance() -> JSON/YAML -> config.instantiate() -> new component
 ```
 
-`instantiate()` calls constructors only. The application remains responsible
+`Config.instantiate()` calls constructors only. The application remains responsible
 for lifecycle methods such as `connect()` and `run()`.
 
 ## Workflow Documents
@@ -32,7 +33,7 @@ for lifecycle methods such as `connect()` and `run()`.
 A workflow document adds command-level structure around components. For
 example, `physicalai run` expects runtime constructor arguments under
 `runtime:` and optional run-method arguments under `run:`. A bare exported
-`RobotRuntime` ComponentConfig is also accepted and reshaped by the loader.
+`RobotRuntime` Config is also accepted and reshaped by the loader.
 
 Nested components retain the same `class_path` + `init_args` form, so a robot,
 action source, model, camera map, and callbacks form one recursive recipe.
@@ -42,10 +43,25 @@ action source, model, camera map, and callbacks form one recursive recipe.
 An inference manifest describes an exported policy package: artifacts,
 features, processors, runners, and compatibility metadata. Its
 `ComponentSpec` supports registry aliases and artifact handling and remains
-separate from strict captured `ComponentConfig` data.
+separate from strict captured `Config` data.
 
 Use workflow configuration for deployment intent. Use manifests for package
 metadata produced during export.
+
+## Generic instantiation (Training and plugins)
+
+Runtime also ships helpers used heavily by Physical AI Studio and
+`jsonargparse`-driven CLIs:
+
+| Entry point | Use when |
+| ----------- | -------- |
+| [`instantiate_obj`](../how-to/config/instantiate-objects.md) | Generic loaders; class chosen by config |
+| [`FromConfig` / `@from_config`](../how-to/config/use-from-config.md) | Class-level `from_yaml`, `from_dict`, `from_config` |
+| [`Config` dataclass subclasses](../how-to/config/instantiate-components.md) | Typed hyperparameter objects with save/load |
+
+Code and tests live under `src/physicalai/config/` and `tests/unit/config/`.
+Studio consumes this package and does not ship a duplicate `physicalai.config`
+tree.
 
 ## Trust
 
--- a/docs/explanation/configuration.md
+++ b/docs/explanation/configuration.md
@@ -53,11 +53,11 @@ metadata produced during export.
 Runtime also ships helpers used heavily by Physical AI Studio and
 `jsonargparse`-driven CLIs:
 
-| Entry point | Use when |
-| ----------- | -------- |
-| [`instantiate_obj`](../how-to/config/instantiate-objects.md) | Generic loaders; class chosen by config |
-| [`FromConfig` / `@from_config`](../how-to/config/use-from-config.md) | Class-level `from_yaml`, `from_dict`, `from_config` |
-| [`Config` dataclass subclasses](../how-to/config/instantiate-components.md) | Typed hyperparameter objects with save/load |
+| Entry point                                                                 | Use when                                            |
+| --------------------------------------------------------------------------- | --------------------------------------------------- |
+| [`instantiate_obj`](../how-to/config/instantiate-objects.md)                | Generic loaders; class chosen by config             |
+| [`FromConfig` / `@from_config`](../how-to/config/use-from-config.md)        | Class-level `from_yaml`, `from_dict`, `from_config` |
+| [`Config` dataclass subclasses](../how-to/config/instantiate-components.md) | Typed hyperparameter objects with save/load         |
 
 Code and tests live under `src/physicalai/config/` and `tests/unit/config/`.
 Studio consumes this package and does not ship a duplicate `physicalai.config`
--- a/docs/explanation/configuration.md
+++ b/docs/explanation/configuration.md
@@ -59,9 +59,10 @@ Runtime also ships helpers used heavily by Physical AI Studio and
 | [`FromConfig` / `@from_config`](../how-to/config/use-from-config.md)        | Class-level `from_yaml`, `from_dict`, `from_config` |
 | [`Config` dataclass subclasses](../how-to/config/instantiate-components.md) | Typed hyperparameter objects with save/load         |
 
-Code and tests live under `src/physicalai/config/` and `tests/unit/config/`.
-Studio consumes this package and does not ship a duplicate `physicalai.config`
-tree.
+Strict captured recipes use the package-level :func:`instantiate`; generic
+loaders use the package-level :func:`instantiate_obj`. Both are intentionally
+available from one public import surface. Generic loader helpers are
+implemented in `physicalai.config.loading`.
 
 ## Trust
 
--- a/docs/how-to/runtime/share-a-robot.md
+++ b/docs/how-to/runtime/share-a-robot.md
@@ -32,7 +32,8 @@ driver = SO101(
     port="/dev/ttyUSB0",
     calibration="~/.cache/calibration/so101.json",  # path stays relative/as given
 )
-robot = SharedRobot.from_config(Config.from_instance(driver), name="left-arm")
+robot = SharedRobot.from_config(driver, name="left-arm")
+# equivalent: SharedRobot.from_config(Config.from_instance(driver), name="left-arm")
 # or: SharedRobot("left-arm", robot={"class_path": "physicalai.robot.SO101", "init_args": {...}})
 robot.connect()
 
--- a/docs/how-to/runtime/share-a-robot.md
+++ b/docs/how-to/runtime/share-a-robot.md
@@ -25,14 +25,14 @@ directly to the constructor or exported explicitly:
 
 ```python
 import numpy as np
-from physicalai.config import to_config
+from physicalai.config import Config
 from physicalai.robot import SO101, SharedRobot
 
 driver = SO101(
     port="/dev/ttyUSB0",
     calibration="~/.cache/calibration/so101.json",  # path stays relative/as given
 )
-robot = SharedRobot.from_config(to_config(driver), name="left-arm")
+robot = SharedRobot.from_config(Config.from_instance(driver), name="left-arm")
 # or: SharedRobot("left-arm", robot={"class_path": "physicalai.robot.SO101", "init_args": {...}})
 robot.connect()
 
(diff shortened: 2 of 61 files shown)


docs/explanation/configuration.md
```diff
@@ -42,11 +43,26 @@ action source, model, camera map, and callbacks form one recursive recipe.
...
+
+| Entry point                                                                 | Use when                                            |
+| --------------------------------------------------------------------------- | --------------------------------------------------- |
+| [`instantiate_obj`](../how-to/config/instantiate-objects.md)                | Generic loaders; class chosen by config             |
+| [`FromConfig` / `@from_config`](../how-to/config/use-from-config.md)        | Class-level `from_yaml`, `from_dict`, `from_config` |
+| [`Config` dataclass subclasses](../how-to/config/instantiate-components.md) | Typed hyperparameter objects with save/load         |
+
+Code and tests live under `src/physicalai/config/` and `tests/unit/config/`.
+Studio consumes this package and does not ship a duplicate `physicalai.config`
+tree.
```
maxxgx: I don't think there is a need to mention this

docs/how-to/runtime/share-a-robot.md
```diff
@@ -25,14 +25,14 @@ directly to the constructor or exported explicitly:
...
-from physicalai.config import to_config
+from physicalai.config import Config
 from physicalai.robot import SO101, SharedRobot
 
 driver = SO101(
     port="/dev/ttyUSB0",
     calibration="~/.cache/calibration/so101.json",  # path stays relative/as given
 )
-robot = SharedRobot.from_config(to_config(driver), name="left-arm")
+robot = SharedRobot.from_config(Config.from_instance(driver), name="left-arm")
```
maxxgx: Would `SharedRobot.from_config(Config(driver))` be possible?

src/physicalai/config/loading.py
maxxgx, line 1: is there any chance of conflicts between this `instantiate.py` and existing `_instantiate.py`?

src/physicalai/runtime/core.py
```diff
@@ -217,20 +217,26 @@ def from_config(cls, config: str | Path) -> Self:
...
+        import yaml  # noqa: PLC0415
         from jsonargparse import ArgumentParser  # noqa: PLC0415
 
-        from physicalai.config import load_yaml  # noqa: PLC0415
-
-        document = _unwrap_runtime_document(load_yaml(config), target=cls)
+        document = yaml.safe_load(Path(config).read_text(encoding="utf-8"))
+        if not isinstance(document, dict):
+            msg = f"runtime config must be a mapping, got {type(document).__name__}"
+            raise TypeError(msg)
```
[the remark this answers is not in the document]
samet-akcay: Addressed in c09ce93eb40e43116de9397efad4149417ca84b4: treat `yaml.safe_load()` `None` as `{}` before validation so empty files fail via jsonargparse / `ConfigError` (missing `class_path`) instead of `TypeError`. Added `test_load_empty_file_as_empty_mapping` and `test_empty_yaml_fails_via_parser_not_type_error`.

src/physicalai/config/_yaml.py
```diff
@@ -70,10 +71,10 @@ def load_yaml(path: str | Path) -> dict[str, JsonValue]:
...
-        ComponentConfigError: If the document is not a mapping.
+        ConfigError: If the document is not a mapping.
     """
     loaded = yaml.safe_load(Path(path).read_text(encoding="utf-8"))
     if not isinstance(loaded, dict):
         msg = f"{str(path)!r}: YAML config must be a mapping, got {type(loaded).__name__}"
-        raise ComponentConfigError(msg)
-    return loaded
+        raise ConfigError(msg)
+    return Config.from_dict(normalize_config(loaded))
```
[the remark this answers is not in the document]
samet-akcay: Addressed in c09ce93eb40e43116de9397efad4149417ca84b4: `load_yaml` now coerces empty documents to `{}`, matching `Config.load` and `instantiate_obj_from_file`. Covered by `test_load_empty_file_as_empty_mapping`.

src/physicalai/robot/transport/_shared_robot.py
```diff
@@ -223,7 +222,7 @@ def __init__(
         self,
         name: str,
         *,
-        robot: ComponentConfig | Mapping[str, object] | None = None,
+        robot: Config | Mapping[str, object] | None = None,
```
maxxgx: Previously, the config was a `TypedDict` which jsonargparse knows how to interpret. But that `Config` is a normal class, it will fall back on the type `Mapping[str, object]`, and jsonargparse doesn't know how to write `object`. Using the `--print_config` command, I'm getting a warning:

```bash
❯ physicalai run --config examples/runtime/runtime.yaml --print_config
/home/max/physicalai/.venv/lib/python3.14/site-packages/jsonargparse/_typehints.py:1776: JsonargparseWarning: 
    Unable to serialize instance physicalai.capture.UVCCamera
...
```

Changing the type to ` Config | Mapping[str, Any] | None` seems to fix this particular issue

src/physicalai/robot/transport/_owner_config.py
maxxgx, line 158: Not this PR's fault, but just realized we need to validate `allow_remote` as well since `allow_remote='no'` would result in `True`

src/physicalai/config/loading.py
```diff
@@ -0,0 +1,186 @@
...
+        return instantiate(Config.from_dict(selected))
+    if target_cls is None:
+        msg = (
+            "Configuration must contain 'class_path' for instantiation, "
+            f"or pass target_cls explicitly. Got keys: {list(selected.keys())}"
+        )
+        raise ValueError(msg)
+    init_args = {name: _instantiate_recursive(value) for name, value in selected.items()}
+    args = init_args.pop("args", ())
+    return target_cls(*args, **init_args)
```
maxxgx, line 111: I think we should document that "args" is treated as positional arguments and therefore is reserved

src/physicalai/config/__init__.py
```diff
@@ -1,60 +1,67 @@
...
+from ._instantiate import instantiate as _strict_instantiate
+from ._normalize import validate_config
+from ._types import ConfigValue, JsonScalar, JsonValue
 from ._yaml import load_yaml, save_yaml, to_yaml
+from .base import Config
 from .importing import import_dotted_path
+from .instantiate import import_class, instantiate_obj
+from .mixin import FromConfig, from_config
+
+instantiate = _strict_instantiate  # ruff: ignore[RUF067]
```
maxxgx, line 49: This is misleading, as 
```python
import physicalai.config.instantiate
```
imports the function and not the module, but 
```python
from physicalai.config.instantiate import instantiate_obj
```
imports from the module and it works.

src/physicalai/config/serializable.py
```diff
@@ -0,0 +1,108 @@
...
+        msg = f"Expected dataclass, got {cls}"
+        raise TypeError(msg)
+    try:
+        hints = get_type_hints(cls)
+    except (NameError, TypeError, AttributeError, KeyError):
+        hints = {}
+    kwargs = {}
+    for field in dataclasses.fields(cls):
+        if field.name in data:
+            kwargs[field.name] = _reconstruct_value(data[field.name], hints.get(field.name, field.type))
```
maxxgx, line 82: If the config contains extra keys, they are ignored silently. Is this intentional?

An issue I see with this approach is that If mistyped an arg then it can be very hard to find out, e.g. using `epoch=5` instead of `epochs=5` and you only find out later that a default value was used instead. I'd prefer an error immediately.

If it's needed for backward compatibility, perhaps we can use a `strict: bool = True` and set to false when needed?

src/physicalai/config/serializable.py
```diff
@@ -0,0 +1,108 @@
...
+            return tuple(value)
+        if len(args) == _VAR_TUPLE_ARG_COUNT and args[1] is ...:
+            return tuple(_reconstruct_value(item, args[0]) for item in value)
+        return tuple(starmap(_reconstruct_value, zip(value, args, strict=False)))
+    actual_type = origin or field_type
+    if isinstance(actual_type, type) and dataclasses.is_dataclass(actual_type) and isinstance(value, dict):
+        return dict_to_dataclass(actual_type, value)
+    if isinstance(actual_type, type) and issubclass(actual_type, Enum) and not isinstance(value, Enum):
+        return actual_type(value)
+    return value
```
maxxgx, line 112:
```suggestion
    if isinstance(actual_type, type) and issubclass(actual_type, Enum) and not isinstance(value, Enum):
        return actual_type(value)
    if isinstance(actual_type, type) and issubclass(actual_type, PurePath) and isinstance(value, str):
        return actual_type(value)
    return value
```

src/physicalai/config/serializable.py
```diff
@@ -0,0 +1,108 @@
...
+        return obj
+    if isinstance(obj, dict):
+        return {(key.value if isinstance(key, Enum) else key): dataclass_to_dict(value) for key, value in obj.items()}
+    if isinstance(obj, (list, tuple)):
+        return [dataclass_to_dict(item) for item in obj]
+    if isinstance(obj, Enum):
+        return obj.value
+    if hasattr(obj, "tolist") and hasattr(obj, "ndim"):
+        return obj.tolist()  # type: ignore[union-attr]
+    return obj
```
maxxgx, line 49: I think a path type check is missing:

```suggestion
    if isinstance(obj, dict):
        return {(key.value if isinstance(key, Enum) else key): dataclass_to_dict(value) for key, value in obj.items()}
    if isinstance(obj, (list, tuple)):
        return [dataclass_to_dict(item) for item in obj]
    if isinstance(obj, Enum):
        return obj.value
    if isinstance(obj, PurePath):
        return str(obj)
    if hasattr(obj, "tolist") and hasattr(obj, "ndim"):
        return obj.tolist()  # type: ignore[union-attr]
    return obj
```

Using a `PurePath` because it catches all path types in python.

See below for the `_reconstruct_value` size

maxxgx, line 49: I think a path type check is missing:

```suggestion
    if isinstance(obj, dict):
        return {(key.value if isinstance(key, Enum) else key): dataclass_to_dict(value) for key, value in obj.items()}
    if isinstance(obj, (list, tuple)):
        return [dataclass_to_dict(item) for item in obj]
    if isinstance(obj, Enum):
        return obj.value
    if isinstance(obj, PurePath):
        return str(obj)
    if hasattr(obj, "tolist") and hasattr(obj, "ndim"):
        return obj.tolist()  # type: ignore[union-attr]
    return obj
```

Using a `PurePath` because it catches all path types in python.

See below for the `_reconstruct_value` size

src/physicalai/config/instantiate.py
```diff
@@ -0,0 +1,186 @@
...
+def _instantiate_recursive(value: object) -> object:
+    if isinstance(value, dict):
+        if "class_path" in value:
+            return Config.from_dict(value).instantiate()
+        return {key: _instantiate_recursive(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [_instantiate_recursive(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_instantiate_recursive(item) for item in value)
+    return value
```
maxxgx: Could we apply `_MAX_CONFIG_DEPTH`? We can increase the value if it's not enough

docs/explanation/configuration.md
samet-akcay: Addressed in 9e7a3b7

docs/how-to/runtime/share-a-robot.md
samet-akcay: Addressed in 9e7a3b7: SharedRobot.from_config now accepts a live @export_config instance and converts it through Config.from_instance().

src/physicalai/config/loading.py
samet-akcay: Addressed in 9e7a3b7: renamed the generic loader module from physicalai.config.instantiate to physicalai.config.loading

src/physicalai/robot/transport/_shared_robot.py
samet-akcay: Addressed in 9e7a3b7: the SharedRobot robot annotation now uses Config | Mapping[str, Any] | None so jsonargparse can serialize the config shape without falling back to object.

src/physicalai/robot/transport/_owner_config.py
samet-akcay: Addressed in c09ce93: owner config parsing now requires allow_remote to be an actual bool before constructing RobotOwnerConfig.

src/physicalai/config/loading.py
samet-akcay: Addressed in 9e7a3b7

src/physicalai/config/__init__.py
samet-akcay: Addressed in 9e7a3b7: generic helpers now live in physicalai.config.loading; the package-level physicalai.config.instantiate remains the strict function without a same-named generic module.

This could work for now, but I think we will need another follow-up PR to improve the hygiene of the API

src/physicalai/config/serializable.py
samet-akcay: Addressed in 9e7a3b7: typed dataclass reconstruction is strict by default and now raises TypeError for unknown keys; callers can explicitly pass strict=False when compatibility is required.

src/physicalai/config/serializable.py
samet-akcay: Addressed in 9e7a3b7

src/physicalai/config/serializable.py
samet-akcay: Addressed in 9e7a3b7

samet-akcay: Addressed in 9e7a3b7

src/physicalai/config/instantiate.py
samet-akcay: Addressed in 9e7a3b7

samet-akcay:
CI fix: the failed code-quality job was caused by a Pyrefly type error in the recursive-depth test fixture (`current["child"]` was inferred as `dict[str, int]`). Fixed in `324c7cc` by explicitly typing the recursive child mapping. `prek run --all-files` now passes locally, and the PR checks are green.

maxxgx (approved):

maxxgx:
Manually tested a few personal yaml configs, everything worked as expected

PR #216, opened 2026-08-04
fix: reject non-finite stats in StatsNormalizer to prevent silent output corruption

AlexanderBarabanov:
**Changes**

This PR fixes a data-integrity issue in `StatsNormalizer` where non-finite values in a
loaded stats file (`-Inf`, `+Inf`, or `NaN`) were silently absorbed by the normalizer
arithmetic and produced plausible-looking finite outputs (`0.0`) instead of propagating the
corruption downstream where it could be detected.

The issue was found by the Atheris fuzzer (harness `stats_normalizer`),
[logs](https://github.com/openvinotoolkit/physicalai/actions/runs/30896023089).

If a loaded stats file is corrupted or replaced with one containing `−Inf` values, the
normalizer currently produces `0.0` for every observation - a plausible-looking zero
tensor that passes any finite-value safety check downstream. 

**Fuzzer logs**

```
=== Uncaught Python exception: ===
AssertionError: StatsNormalizer silently produced finite output despite NaN/Inf stats
(mode='mean_std').  NaN/Inf must propagate to be detectable downstream.
Traceback (most recent call last):
  File "tests/fuzz/harnesses/fuzz_stats_normalizer.py", line 82, in test_one_input
    assert not np.all(np.isfinite(outputs[feature_name])), (
AssertionError: StatsNormalizer silently produced finite output despite NaN/Inf stats

==2468== ERROR: libFuzzer: fuzz target exited
stat::number_of_executed_units: 391809
artifact_prefix='./'; Test unit written to ./crash-b48490b7c314e587560babe80b6d1b29a342a779
Base64: DQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQAAAAAAgP//AA0NDQ0NDQ0N
      DQ0NDQ0NDQ0AAAAAAID//wAE
```

**Fix**: Added an `np.isfinite` guard at the start of each normalization branch in `_normalize()`.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -15,9 +15,6 @@ on:
     paths:
       - "tests/fuzz/**"
       - ".github/workflows/fuzz.yml"
-  push:
-    branches:
-      - fuzz-fix-2
 
 permissions: {}
 
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -15,6 +15,9 @@ on:
     paths:
       - "tests/fuzz/**"
       - ".github/workflows/fuzz.yml"
+  push:
+    branches:
+      - fuzz-fix-2
 
 permissions: {}
 
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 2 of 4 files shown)


alfieroddan (approved):
Thanks!

PR #223, opened 2026-08-06
fix: dotted paths with an empty segment

AlexanderBarabanov:
This PR fixes a fuzzer discovered crash in `import_dotted_path()`: dotted paths with an empty segment (e.g. leading/trailing/consecutive `.`, such as `.foo` or `foo..bar`) were passed straight to `importlib.import_module()`, which raises an uncaught `TypeError` for relative import style names instead of the documented `ValueError`.


ivanzati (dismissed):
Thank you, Alexander!

MarkRedeman (dismissed):

MarkRedeman (approved):

PR #227, opened 2026-08-07
feat(config): use jsonargparse as construction engine

samet-akcay:
## Summary

Minimal configuration architecture built around `Config` and jsonargparse:

- Removed generic utility facades: `loading.py`, `mixin.py`, `serializable.py`, `_yaml.py`, and `_envelope.py`.
- Removed public `instantiate_obj`, `FromConfig`, package-level YAML helpers, import helpers, and generic normalization helpers.
- `Config` directly owns typed mapping/file parsing, saving, portable recipe load/save, exportability, and class resolution.
- Known-type construction delegates to jsonargparse; `Config.instantiate(expected_type=...)` is used at robot/camera ownership boundaries.
- Schema-free construction remains isolated only for direct recipe compatibility.
- Inference uses expected-type jsonargparse construction while retaining domain alias/artifact policy.
- Transport envelope validation moved to the robot/camera modules that own those schemas.
- Adopted jsonargparse-native enum names in persisted typed/exported configs.
- Deleted utility-only tests, docs, and navigation; updated examples and skills to canonical APIs.
- Updated the implementation specification to reflect the reduced source tree and removal policy.


src/physicalai/config/_normalize.py
```diff
@@ -115,7 +115,7 @@ def _try_normalize_scalar(value: object, *, path: str) -> tuple[bool, JsonValue]
 
 
 def _normalize_enum(value: Enum, *, path: str) -> JsonValue:
-    enum_value = value.value
+    enum_value = value.name
```
maxxgx, line 120: This breaks enum object round trip (instance -> config -> instance). For example:
```python
from physicalai.capture import UVCCamera, ColorMode
from physicalai.config import Config

cam = UVCCamera(device="/dev/video0", color_mode=ColorMode.BGR)
config = Config.from_instance(cam)
instance = config.instantiate()
```
results in:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/max/physicalai/src/physicalai/config/base.py", line 213, in instantiate
    return instantiate(self, expected_type=expected_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/max/physicalai/src/physicalai/config/_instantiate.py", line 298, in instantiate
    _preflight_config(raw, path="", depth=0, seen=set())
  File "/home/max/physicalai/src/physicalai/config/_instantiate.py", line 58, in _preflight_config
    _preflight_value(item, path=child_path, depth=depth + 1, seen=seen)
  File "/home/max/physicalai/src/physicalai/config/_instantiate.py", line 126, in _preflight_value
    raise ConfigError(msg)
physicalai.config._errors.ConfigError: physicalai.capture.UVCCamera.init_args.color_mode: Enum is not a JSON-compatible config value; pass its value instead
```

src/physicalai/config/base.py
```diff
@@ -7,18 +7,82 @@
...
+def save_class_config(value: _T, path: str | Path) -> None:
+    """Serialize a typed config through jsonargparse."""
+    parser = ArgumentParser(exit_on_error=False)
+    parser.add_class_arguments(type(value), "object")
+    values = dataclasses.asdict(cast("dataclasses.DataclassInstance", value))
+    namespace = parser.parse_object({"object": values}, defaults=False)
+    dumped = yaml.safe_load(parser.dump(namespace, format="yaml"))
+    if isinstance(dumped, Mapping) and "object" in dumped:
+        dumped = dumped["object"]
+    Path(path).write_text(yaml.safe_dump(dumped, sort_keys=False), encoding="utf-8")
```
maxxgx: I think copilot caught this one too in parse function. 

This changes the format for typed configs:
- main branch, the **outer** class uses `class_path / init_args` while nested use bare fields
- this branch, the outer class uses bare fields while **nested** class uses `class_path / init_args`

This effectively "swaps" the format,  for example:
```yaml
# written by main                     # written by this branch
class_path: mypkg.TypedConfig         nested:
init_args:                              class_path: mypkg.Nested
  nested: {value: 3}                    init_args: {value: 3}
  mode: fast                          mode: FAST
  shape: [2, 4]                       shape: [2, 4]
```

I'm not sure if this create any issues with existing configs. 
I prefer main's format since `.to_jsonargparse()` produces that as well. But I'm also fine with the new version if we make parsing support both formats

src/physicalai/robot/transport/_owner_config.py
```diff
@@ -104,10 +120,13 @@ def coerce_robot_config_input(robot: Config | Mapping[str, object] | object) ->
...
-    if type(robot) is Config or isinstance(robot, Mapping):
+    if type(robot) is Config:
+        return robot
+    if isinstance(robot, Mapping):
         return robot
-    if is_config_exportable(robot):
-        return Config.from_instance(robot)
+    exportable = cast("ConfigExportable", robot)
+    if exportable.supports_config_export():
+        return exportable.as_config()
```
maxxgx: If the robot is not exportable we will get a `AttributeError: object has no attribute 'supports_config_export'` instead of the previous `TypeError`.

I think it's better to check for `isinstance`:
```python
if isinstance(robot, ConfigExportable):
     return robot.as_config()
```

src/physicalai/config/_normalize.py
maxxgx: It would be nice to have more tests to cover round trip behaviour

src/physicalai/config/loading.py
```diff
@@ -133,12 +165,16 @@ def instantiate_obj_from_dataclass(
 ) -> object:
     """Instantiate from a dataclass instance.
 
+    .. deprecated::
+        Use jsonargparse or :func:`~physicalai.config.base.parse_class_config` instead.
+
```
maxxgx, line 170: Are deprecated functions safe to remove?

maxxgx:
Note: description says `Removed generic utility facades: loading.py, mixin.py, serializable.py, _yaml.py, and _envelope.py.`, but they still present (deprecated?). What was the reason to not remove them?

src/physicalai/config/_instantiate.py
```diff
@@ -281,11 +283,32 @@ def instantiate(config: Config | Mapping[str, JsonValue]) -> object:
...
+            msg = f"cannot import class_path {class_path!r}: {exc}"
+            raise ConfigImportError(msg) from exc
+        if not isinstance(target, type):
+            msg = f"{class_path!r} does not resolve to a class"
+            raise ConfigImportError(msg)
+        result = parse_class_config(target, cast("Mapping[str, object]", raw["init_args"]))
+        if not isinstance(result, expected_type):
+            msg = f"{class_path!r} does not satisfy {expected_type.__name__}"
+            raise ConfigError(msg)
+        return result
```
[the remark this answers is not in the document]
samet-akcay: Addressed in 66cb212.

src/physicalai/config/base.py
```diff
@@ -1,24 +1,89 @@
...
+    """
+    data = yaml.safe_load(Path(path).read_text(encoding="utf-8")) or {}
+    if not isinstance(data, Mapping):
+        msg = f"Expected configuration root to be a mapping, got {type(data).__name__}"
+        raise TypeError(msg)
+    values = data.get("object", data)
+    if not isinstance(values, Mapping):
+        msg = "Expected configuration values to be a mapping"
+        raise TypeError(msg)
+    return parse_class_config(target, values, defaults=defaults)
```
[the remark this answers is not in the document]
samet-akcay: Addressed in d41c65a.

src/physicalai/robot/transport/_owner_config.py
```diff
@@ -72,27 +70,44 @@ def validate_owner_config(data: Mapping[str, Any]) -> Config:
...
-    )
+    recipe = robot if isinstance(robot, Config) else Config.from_dict(robot)
+    if "." not in recipe.class_path or not recipe.class_path.strip():
+        msg = f"robot_class must be a nonempty dotted path, got {recipe.class_path!r}"
+        raise ValueError(msg)
+    try:
+        json.dumps(recipe.to_dict())
+    except (TypeError, ValueError) as exc:
+        msg = "robot config must be JSON-serializable (e.g. paths as str, not objects)"
+        raise ValueError(msg) from exc
```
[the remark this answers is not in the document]
samet-akcay: Addressed in 66cb212.

tests/unit/config/test_export_config.py
```diff
@@ -24,12 +24,15 @@
...
-    to_config,
 )
+from physicalai.config.importing import import_dotted_path
+
+to_config = Config.from_instance
+is_config_exportable = Config.is_exportable
+
+
+def instantiate(value: object) -> object:
+    return Config.from_dict(cast("Mapping[str, object]", value)).instantiate()
```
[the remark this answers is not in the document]
samet-akcay: Addressed in 66cb212.

src/physicalai/config/_normalize.py
samet-akcay: Addressed in 9c5b7b1.

src/physicalai/config/base.py
samet-akcay: Addressed in d41c65a.

src/physicalai/robot/transport/_owner_config.py
samet-akcay: Addressed in 0f635b0.

src/physicalai/config/_normalize.py
samet-akcay: Addressed in 9c5b7b1 and 66cb212.

src/physicalai/config/loading.py
samet-akcay: Not removing in this PR: deprecated shims remain until Studio migrates. Follow-up Runtime PR after Studio PR2.

src/physicalai/config/base.py
maxxgx: Looks like the enum fix only handles the direct `Config.instantiate()` path, but not typed `Config` subclasses parsed through jsonargparse.

Configs written by `main` store enum values, for example:

```yaml
class_path: mypkg.TypedConfig
init_args:
  mode: fast
```
where:
```python
class Mode(Enum):
    FAST = "fast"
```
Loading this on the PR branch, `TypedConfig.load(path)` raises:
```
ArgumentError: Parser key "object.mode":
  Expected a member of <enum 'Mode'>: FAST. Got value: fast
```
The new tests cover mode: `FAST`, but not the current mode: `fast`.

Could we make typed config loading accept both enum values and names? I think we need this for backwards compatibility

samet-akcay: yeah that's fair

samet-akcay: Added both support in https://github.com/openvinotoolkit/physicalai/pull/227/changes/3687ee9c38b9be729ae09baa4bdd253333a1e4d7, and added a simple test. Perhaps you could check with your config ?

maxxgx: Thanks, they are working now

maxxgx (approved):
LGTM, manual config roundtrip passed

PR #230, opened 2026-08-10
feat: add IP camera support

maxxgx:
## Summary

Adds IP camera backend via opencv.

## Validation

- Tested on live RTSP IP camera

## Breaking changes

- None

## Related issues

- Closes #

### DIFF
--- a/src/physicalai/capture/cameras/ip/_camera.py
+++ b/src/physicalai/capture/cameras/ip/_camera.py
@@ -1,0 +1,302 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import contextlib
+import os
+import threading
+import time
+from typing import TYPE_CHECKING
+from urllib.parse import urlsplit, urlunsplit
+
+import cv2
+from loguru import logger
+
+from physicalai.capture.camera import Camera, ColorMode
+from physicalai.capture.errors import CaptureError, CaptureTimeoutError, NotConnectedError
+from physicalai.capture.frame import Frame
+from physicalai.config import export_config
+
+if TYPE_CHECKING:
+    import numpy as np
+
+# OpenCV's Python bindings expose no CAP_PROP for RTSP transport selection,
+# so this FFmpeg-backend env var (read fresh on each VideoCapture open) is
+# the only portable way to force TCP: FFmpeg's UDP default is often silently
+# dropped by routers/NAT, leaving the RTSP session "open" but frame-less
+# forever. `stimeout` (microseconds) is a session-wide socket timeout — a
+# real, working backstop against a stalled socket, unlike
+# CAP_PROP_READ_TIMEOUT_MSEC (see read()/_wait_first_frame()). Guarded by a
+# lock since this is process-global state and each connect() computes its
+# own stimeout value from its own `timeout`.
+_RTSP_TRANSPORT_LOCK = threading.Lock()
+
+
+def _rtsp_capture_options(timeout: float) -> str:
+    """Build the OPENCV_FFMPEG_CAPTURE_OPTIONS value for an RTSP connect().
+
+    Returns:
+        A pipe-separated ``key;value`` option string for FFmpeg's rtsp demuxer.
+    """
+    stimeout_us = int(timeout * 1_000_000)
+    return f"rtsp_transport;tcp|stimeout;{stimeout_us}"
+
+
+def _redact_url(url: str) -> str:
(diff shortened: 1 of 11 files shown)


src/physicalai/capture/cameras/ip/_camera.py
```diff
@@ -0,0 +1,288 @@
...
+                cap.set(cv2.CAP_PROP_READ_TIMEOUT_MSEC, remaining * 1000)
+                ok, frame = cap.read()
+                if ok and frame is not None:
+                    data = self._convert(frame)
+                    self._height, self._width = data.shape[:2]
+                    self._last_frame_data = data
+                    self._connected = True
+                    self._sequence = 0
+                    logger.info(f"IP camera {_redact_url(self._url)} connected ({self._width}x{self._height})")
+                    return
```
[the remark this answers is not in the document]
maxxgx: Agreed — confirmed via a test that reproduces it: calling `read_latest()` right after `connect()`, before any live `read()`/`read_latest()` has succeeded, returned `Frame(timestamp=0.0, sequence=0)` from the cached fallback instead of the actual capture time. Fixed in bd394976 by setting `self._last_timestamp = time.monotonic()` alongside `self._last_frame_data` in `_wait_first_frame()`. Added `test_read_latest_falls_back_to_connect_time_frame_with_valid_timestamp`, which I verified fails without the fix and passes with it.

src/physicalai/capture/transport/_spec.py
```diff
@@ -149,7 +172,12 @@ def derive_service_name(
     init_args = camera.get("init_args", {})
     if not isinstance(init_args, Mapping):
         init_args = {}
-    device_id = init_args.get("serial_number", init_args.get("device", 0))
+    device_id = init_args.get("serial_number", init_args.get("device"))
+    if device_id is None:
```
MarkRedeman, line 176: Perhaps we can also add a `device_id` method to a camera?
Edit: it looks like the ip camera does have  device id, can't we use it?

src/physicalai/capture/cameras/ip/_camera.py
```diff
@@ -0,0 +1,302 @@
...
+    """Build the OPENCV_FFMPEG_CAPTURE_OPTIONS value for an RTSP connect().
+
+    Returns:
+        A pipe-separated ``key;value`` option string for FFmpeg's rtsp demuxer.
+    """
+    stimeout_us = int(timeout * 1_000_000)
+    return f"rtsp_transport;tcp|stimeout;{stimeout_us}"
+
+
+def _redact_url(url: str) -> str:
```
MarkRedeman, line 46: This looks quite similar to `_url_token`, reuse on of these?

src/physicalai/capture/cameras/ip/_camera.py
```diff
@@ -0,0 +1,302 @@
...
+        try:
+            # Only CAP_PROP_OPEN_TIMEOUT_MSEC goes here — this build rejects
+            # the whole params array if CAP_PROP_READ_TIMEOUT_MSEC/BUFFERSIZE
+            # are also included (verified against real opencv-python-headless,
+            # not reachable via the mocked tests). Both are applied via
+            # cap.set() below instead, after a successful open.
+            with _RTSP_TRANSPORT_LOCK if is_rtsp else contextlib.nullcontext():
+                previous_options = os.environ.get("OPENCV_FFMPEG_CAPTURE_OPTIONS") if is_rtsp else None
+                if is_rtsp:
+                    os.environ["OPENCV_FFMPEG_CAPTURE_OPTIONS"] = _rtsp_capture_options(timeout)
```
MarkRedeman, line 131: What if there are multiple ip cameras, wouldn't they overwrite the same environment variables?

src/physicalai/capture/transport/_spec.py
maxxgx: We do have a `device_id` property but it's constructed by a camera object, while here we only have a config

src/physicalai/capture/cameras/ip/_camera.py
maxxgx: Yeah they quite similar but slightly different semantics, `_url_token` is more restrictive while `_redact_url` only strips credentials. I don't think it's worth to create a helper function for these 3 lines

src/physicalai/capture/cameras/ip/_camera.py
maxxgx: `_RTSP_TRANSPORT_LOCK` should handle race conditions. The env var is only used by opencv when opening the stream

MarkRedeman (dismissed):

MarkRedeman (approved):

PR #231, opened 2026-08-10
fix: add image_key_reorder_map and num_cameras to ResizeSmolVLA

pwolnows:
## Problem

`physical-ai-studio` exports `image_key_reorder_map` and `num_cameras` into the `ComponentSpec` manifest (`policy_preprocessor.json`) for SmolVLA policies. On the physicalai inference side, `instantiate_component` forwards all `flat_params` from the spec as `**kwargs` to the registered class constructor. Since `ResizeSmolVLA.__init__` did not accept these parameters, loading any SmolVLA model exported with a non-trivial camera configuration raised:

```
TypeError: ResizeSmolVLA.__init__() got an unexpected keyword argument 'image_key_reorder_map'
```

This is a cross-repo desync introduced when `physical-ai-studio` added camera reordering support without a corresponding change in `physicalai`.

### `src/physicalai/inference/preprocessors/smolvla.py`

- Added `image_key_reorder_map: dict[str, int] | None = None` and `num_cameras: int = 0` parameters to `__init__`
- Added `_resolve_image_order(img_keys)` helper that:
  - Without a map: preserves natural key order
  - With a map: sorts keys by their assigned camera slot index
  - With `num_cameras > 0`: allocates a fixed-length layout, filling unoccupied slots with `None`
- `None` slots in the image list are filled with a black image (`-1.0` after normalisation) and a zero mask, matching the behaviour expected by the SmolVLA model for missing cameras

### `tests/unit/inference/preprocessors/test_smolvla.py`

Added four new test classes:
- **`TestResizeSmolVLAImageKeyReorderMapInit`** — constructor accepts new params; `None` map normalises to `{}`
- **`TestResizeSmolVLAResolveImageOrder`** — unit tests for `_resolve_image_order`: natural order, map-driven reorder, `num_cameras` None-slot filling
- **`TestResizeSmolVLANoneCameraSlot`** — end-to-end: missing camera slot → black image + zero mask
- **`TestResizeSmolVLAComponentSpecCompat`** — regression test that exercises the exact code path that was broken: `instantiate_component(ComponentSpec(type="smolvla_resize", image_key_reorder_map=..., num_cameras=...))` must succeed and produce a correctly configured instance


src/physicalai/inference/preprocessors/smolvla.py
```diff
@@ -25,15 +25,27 @@ class ResizeSmolVLA(Preprocessor):
             as (height, width). Defaults to (512, 512).
     """
 
-    def __init__(self, image_resolution: tuple[int, int] = (512, 512)) -> None:
-        """Initialize the SmolVLA numpy-based preprocessor.
-
-        Args:
-            image_resolution (tuple[int, int]): The target resolution for input images
-                as (height, width). Defaults to (512, 512).
-        """
```
maxxgx, line 34: Can you add updated docstring back?

src/physicalai/inference/preprocessors/smolvla.py
```diff
@@ -25,15 +25,27 @@ class ResizeSmolVLA(Preprocessor):
...
+    def __init__(
+        self,
+        image_resolution: tuple[int, int] = (512, 512),
+        image_key_reorder_map: dict[str, int] | None = None,
+        num_cameras: int = 0,
+    ) -> None:
         super().__init__()
         self.image_resolution = image_resolution
+        # image_key_reorder_map maps input key → camera slot index for reordering
+        self._image_key_reorder_map: dict[str, int] = image_key_reorder_map or {}
```
maxxgx: `self._image_key_reorder_map` can be kept optional instead of `{}` since the empty value is never used anyway

maxxgx: Might need to add a `images.` prefix to the keys if Studio exports bare keys, e.g. `gripper` vs `images.gripper`.

src/physicalai/inference/preprocessors/smolvla.py
```diff
@@ -62,15 +74,27 @@ def __call__(self, inputs: dict[str, np.ndarray]) -> dict[str, np.ndarray]:
...
+            ordered = self._resolve_image_order(img_keys)
+            images = [inputs[IMAGES][k] if k is not None else None for k in ordered]
         else:
             img_keys = [key for key in inputs if key.startswith(IMAGES)]
-            images = [inputs[img_keys[0]]] if len(img_keys) == 1 else [inputs[key] for key in img_keys]
+            if len(img_keys) == 1:
+                images = [inputs[img_keys[0]]]
+            else:
+                ordered = self._resolve_image_order(img_keys)
+                images = [inputs[k] if k is not None else None for k in ordered]
```
maxxgx, line 132: I think when there is a single camera, PAI expects a dict with a single key: `images`. Would this still work?

cc @sovrasov

src/physicalai/inference/preprocessors/smolvla.py
pwolnows: Good catch - the concern was real and is already addressed with _bare_key() helper normalises both sides to bare keys at __init__ and _resolve_image_order time. 
So whether Studio exports {"wrist": 0} (bare) or {"images.wrist": 0} (prefixed), and whether the runtime flat-key path sends "images.wrist" or "wrist", they all match against the same bare-key representation internally. 
No prefix addition is needed in either direction.

src/physicalai/inference/preprocessors/smolvla.py
pwolnows: Restored

src/physicalai/inference/preprocessors/smolvla.py
pwolnows: Yes, the single-camera case still works. When the input is {"images": ndarray} or {"images.cam": ndarray}, the preprocessor takes the first branch at line 128
Added test_single_camera_ndarray_bypasses_reorder_map and test_single_flat_key_bypasses_reorder_map to cover these cases explicitly.

sovrasov (approved):

PR #232, opened 2026-08-11
fix(ci): exclude Lob detector from secret scan workflow

AlexanderBarabanov:
This PR removes `Lob` detector to suppress false positives ([example](https://github.com/openvinotoolkit/physicalai/actions/runs/31478179606)) as this repo does not use the Lob postal API.


ivanzati (approved):
Thank you, Alexander!

PR #234, opened 2026-08-12
feat: MolmoAct2 processors

alfieroddan:
## Summary

Corresponding PR to introduce numpy processors 

https://github.com/open-edge-platform/physical-ai-studio/pull/809

Other changes: we've adjusted the stats normalization to include the new mask parameter.


src/physicalai/inference/preprocessors/molmoact2.py
```diff
@@ -0,0 +1,304 @@
...
+        self,
+        signs: list[float] | None = None,
+        offsets: list[float] | None = None,
+    ) -> None:
+        """Use Studio SO101 defaults unless compatible overrides are supplied.
+
+        Raises:
+            ValueError: If signs and offsets have different lengths.
+        """
+        signs = list(SO101_JOINT_SIGNS) if signs is None else signs
```
kdijkstraintel: Why use offsets of this specific embodiment (so101) by default? Introducing a new embodiment in the future might result in using wrong default offsets.

src/physicalai/inference/preprocessors/molmoact2.py
```diff
@@ -0,0 +1,304 @@
...
+            msg = f"num_state_tokens must be > 0, got {num_state_tokens}."
+            raise ValueError(msg)
+        self.image_keys = list(image_keys)
+        self.image_size = tuple(image_size)
+        self.num_state_tokens = num_state_tokens
+        self.setup_type = setup_type
+        self.control_mode = control_mode
+        self.add_setup_tokens = add_setup_tokens
+        self.add_control_tokens = add_control_tokens
+        self.joint_transform = JointFrameTransform(joint_signs, joint_offsets) if adapt_to_so101 else None
```
kdijkstraintel: Why use a flag for a specific embodiment (adapt_to_so101)? Why not generalize and pass a JointFrameTransform() to this constructor so that the parent can initialize it this depending on the specific embodiement used?

kdijkstraintel (changes requested):

src/physicalai/inference/preprocessors/molmoact2.py
alfieroddan: https://github.com/openvinotoolkit/physicalai/pull/234/commits/868ea1129d5620b78dcffd3c31d0aa0fd0eb4444

src/physicalai/inference/preprocessors/molmoact2.py
alfieroddan: https://github.com/openvinotoolkit/physicalai/pull/234/commits/868ea1129d5620b78dcffd3c31d0aa0fd0eb4444

src/physicalai/inference/postprocessors/molmoact2.py
```diff
@@ -0,0 +1,62 @@
...
+        """Postprocess the model action output.
+
+        Returns:
+            Outputs with the canonical denormalized action.
+
+        Raises:
+            ValueError: If no action output is present.
+        """
+        result = dict(outputs)
+        action = result.get(ACTION, result.get("actions"))
```
sovrasov: Can PAS or OV inference create a dict with `actions` key?
Which entity controls the key name?

src/physicalai/inference/joint_transform.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2026 Intel Corporation
```
sovrasov: Looks like a particular transform is too low-level to be located right in the `inference` namespace, which is also shared with pas-train. Is it utilized outside of the corresponding postprocessor? Maybe move it there?

alfieroddan: Moved to postprocessor in https://github.com/openvinotoolkit/physicalai/pull/234/commits/628cf1c1411b63fe315ad9dbd16b89df34689bc7

src/physicalai/inference/postprocessors/molmoact2.py
alfieroddan: Addressed in https://github.com/openvinotoolkit/physicalai/pull/234/commits/5ab81540fc8d1ad80e0821acd4b1ee5008904c9c

sovrasov (dismissed):

kdijkstraintel (dismissed):
Looks good.

samet-akcay:
would it be an idea to re-structure src/physicalai/inference/postprocessors directory as it is getting a bit cluttered? Maybe a subpackage per-model? Something like

 ```text
   src/physicalai/inference/preprocessors/
   ├── __init__.py
   ├── base.py
   ├── stats_normalizer.py
   ├── resize.py
   ├── pi05.py
   ├── rldx1/
   │   ├── __init__.py
   │   ├── processor.py
   │   ├── rope.py
   │   └── token_composer.py
   └── molmoact2/
       ├── __init__.py
       ├── processor.py
       ├── inputs.py
       └── image.py
 ```

__init__.py could expose the classes:

 ```python
   from physicalai.inference.preprocessors.molmoact2.inputs import MolmoAct2Inputs
   from physicalai.inference.preprocessors.molmoact2.processor import MolmoAct2Preprocessor

   __all__ = ["MolmoAct2Inputs", "MolmoAct2Preprocessor"]
 ```

so relative imports would still work?

samet-akcay (approved):

PR #237, opened 2026-08-12
feat: add policy source reset

maxxgx:
## Summary

Makes the runtime reusable across runs and policy episodes for long-lived Studio sessions:

  • PolicySource.reset() — discard queued/stale actions without stopping the loop or restarting the inference worker
  • PolicySource.warmup() — optional eager first inference so arming policy doesn't stall a tick
  • Callback lifecycle fix — run() flushes callbacks; disconnect() closes them (fixes silent telemetry loss on second run())
  • Execution reset support — sync/async/RTC invalidate in-flight work via _episode_id and support safe episode boundaries
  • Docs + tests — API docs updated; ~900 lines of new unit tests

For more context: https://github.com/open-edge-platform/physical-ai-studio/pull/936/changes#diff-64daf1c928767ba3233bd70eaf2285c62b9c7323e6de4fb929cc40a29e67de7d

## Why

 Studio runs one long session and switches between hold, teleop, and policy. Stopping a task and restarting later must not replay old queued actions or require tearing down the whole runtime (which closes dataset writers and restarts inference workers). Callbacks also need to survive consecutive run() calls, previously JsonlCallback / AsyncCallback broke silently on the second run.

### DIFF
--- a/src/physicalai/runtime/action_sources/policy.py
+++ b/src/physicalai/runtime/action_sources/policy.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import threading
 import time
 from typing import TYPE_CHECKING, Any
 
@@ -14,6 +15,7 @@ from physicalai.config import export_config
 from physicalai.inference.constants import IMAGES, STATE, TASK
 from physicalai.runtime.action_sources.base import ActionSource
 from physicalai.runtime.events import MetricsEvent
+from physicalai.runtime.execution.base import Execution
 from physicalai.runtime.execution.queue import ChunkedActionQueue
 from physicalai.runtime.execution.sync import SyncExecution
 from physicalai.runtime.smoothers import LerpSmoother
@@ -25,7 +27,6 @@ if TYPE_CHECKING:
     from physicalai.inference.model import InferenceModel
     from physicalai.robot.interface import RobotObservation
     from physicalai.runtime._callback_bus import _CallbackBus
-    from physicalai.runtime.execution.base import Execution
     from physicalai.runtime.execution.queue import ActionQueue
 
 _DEFAULT_LERP_FRAMES = 5
@@ -55,10 +56,12 @@ class PolicySource(ActionSource):
         self._bus: _CallbackBus | None = None
         self._session_id: str = ""
         self._connected = False
+        self._state_lock = threading.RLock()
 
     def set_task(self, task: str | None) -> None:
         """Update the task string used for the *next* inference request."""
-        self._task = task
+        with self._state_lock:
+            self._task = task
 
     @property
     def action_queue(self) -> ActionQueue:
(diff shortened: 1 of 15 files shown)


src/physicalai/runtime/execution/async_execution.py
```diff
@@ -245,17 +279,23 @@ def _run(self, stop_event: threading.Event, obs_ready: threading.Event) -> None:
...
 
                 if self._model is None or self._queue is None:
                     raise RuntimeError(NOT_STARTED)  # noqa: TRY301
                 t0 = time.perf_counter()
-                actions = self._model.predict_action_chunk(obs)
-                latency = time.perf_counter() - t0
+                with self._model_lock:
+                    with self._lock:
+                        if episode_id != self._episode_id:
+                            self._running_inference = False
```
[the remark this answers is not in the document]
maxxgx: False positive, locks released before blocking

src/physicalai/runtime/action_sources/policy.py
```diff
@@ -71,18 +74,75 @@ def action_queue(self) -> ActionQueue:
...
+                raise RuntimeError(msg)
+            if self._warmed_up:
+                return
+            self._execution.warmup(observation)
+            self._warmed_up = True
+
+    @property
+    def _supports_reset(self) -> bool:
+        reset = getattr(type(self._execution), "reset", None)
+        return reset is not None and reset is not Execution.reset
```
MarkRedeman, line 145: I don't get this line, why the `reset is not Execution.reset` check?

src/physicalai/runtime/execution/async_execution.py
```diff
@@ -62,6 +63,11 @@ def __init__(
         self._thread: threading.Thread | None = None
         self._death_cause: BaseException | None = None
         self._inference_count: int = 0
+        # Episode counter bumped by start()/reset(). Each observation handed to
+        # the worker carries the current value; if reset() lands while inference
+        # is in flight, the worker's episode_id no longer matches and the result
+        # is discarded instead of reaching the queue.
+        self._episode_id: int = 0
```
MarkRedeman, line 70: I'm wondering if `episode_id` should really be an execution thing. Suppose we use a `TeleopSource` then the runtime should also keep track of an `episode_id` if we were to use this for recording episodes.
Would moving this to the runtime be more appropriate?

MarkRedeman (approved):
I think overall the PR looks logical, but it is quite hard to follow all the changes.
It's not entirely clear to me if we need this many locks just because of the complexity of the runtime, or because maybe we are missing some simplifications?

One thing I'd want to look at is if `episode_id` should be moved into the runtime instead of "hidden" in an execution class that is specific to inference policies: teleop actions may also need to keep track of the episode a user is recording.

For later: when we add dataset recording with human in the loop we should also be able to keep track of the "sources" of an action. This is important as it allows us to distinguish when a recording's actions are coming from policy inference or telop.

src/physicalai/runtime/action_sources/policy.py
maxxgx: Serves to make reset optional for future policy sources. Could make it mandatory but I thought it would be better to have as optional and see how it would work with Studio before committing to a new interface

src/physicalai/runtime/execution/async_execution.py
maxxgx: I think "episode_id" is slightly misleading here as it's not the recording episode. It serves to differentiate between different resets so that it can discard the action queue. Open for a clearer name if you have one

MarkRedeman: Ah makes sense. My AI was suggesting `generation_id` or `reset_generation_id`, which I think can make sense.

maxxgx: Will do the renaming in a followup, I was thinking "incarnation" could be a good fit. 

> incarnation refers to a distinct, sequential instance, version, or life-cycle state of an entity—such as a process, node, or database—following a reset, reboot, or recovery. It is frequently tracked using an incarnation number to prevent old or delayed network messages from a previous failure from interfering with the current active state.

maxxgx: followup: https://github.com/openvinotoolkit/physicalai/pull/238

PR #253, opened 2026-09-02
fix(config): honor strict=False in Config.from_dict

### LINKED ISSUE #251: Config.from_dict(strict=False) still rejects unknown keys (regression since a7ee74d0)
### Description

`Config.from_dict(data, strict=False)` on a typed dataclass `Config` subclass is documented to "reject unknown keys" only when `strict=True`, implying `strict=False` should tolerate extra keys in `data`. In the current implementation this is not the case: `from_dict` still forwards the *entire, unfiltered* `data` mapping to `parse_class_config`, which builds a jsonargparse `ArgumentParser` containing only the dataclass's known fields and calls `parser.parse_object(...)`. jsonargparse itself has no notion of `strict=False` and raises on any key without a matching argument, so `strict=False` has become a no-op that can never succeed where `strict=True` also fails.

```python
# physicalai/config/base.py
@classmethod
def from_dict(cls, data, *, strict=True):
    ...
    if strict:
        field_names = {field.name for field in dataclasses.fields(cls)}
        extras = set(data) - field_names
        if extras:
            raise TypeError(...)
    return parse_class_config(cls, data)  # <-- `data` is never filtered when strict=False
```

### Regression

This behavior changed between two versions vendored by `physical-ai-studio`:

- `9be95d88` (older): `Config.from_dict` used `dict_to_dataclass`, which iterates only over the dataclass's own fields and simply ignores any keys not present — `strict=False` worked as documented.
- `a7ee74d0` (current, introduced in physical-ai-studio#1024 / "Upgrade to dictionary based payload"): `Config.from_dict` switched to `parse_class_config` (jsonargparse-based), which rejects unknown keys unconditionally.

### Reproduction

```python
import dataclasses
from physicalai.config import Config

@dataclasses.dataclass(frozen=True)
class Dummy(Config):
    a: int = 1
    b: str = "x"

data = {
    "a": 2,
    "b": "y",
    "input_features": {"observation.images.base_0_rgb": {"type": "VISUAL"}},
}

Dummy.from_dict(data, strict=False)
# jsonargparse.ArgumentError: Group 'object' does not accept option
# 'input_features.observation.images.base_0_rgb.type'
```

### Impact

Any consumer relying on `strict=False` to tolerate extra keys breaks. Concretely, this caused physical-ai-studio#1069: loading any pretrained `lerobot`-format `config.json` (e.g. `lerobot/pi05_base`) via `Pi05Config.from_dict(hf_config, strict=False)` failed, because lerobot config files always carry keys like `input_features`/`output_features`/`repo_id`/`license` that have no matching dataclass field. The same pattern affects `Rldx1Config` and `SmolVLAConfig`.

Studio worked around this downstream by pre-filtering `data` to known dataclass fields before calling `from_dict` (see `physicalai.policies.utils.pretrained.known_config_fields_only` in physical-ai-studio), but that workaround should not be necessary given the documented `strict` contract.

### Expected Behavior

`Config.from_dict(data, strict=False)` should filter `data` down to the dataclass's known fields (as the pre-a7ee74d0 implementation did) before delegating to the jsonargparse-based parser, so unknown keys are silently dropped rather than raising.

### Suggested Fix

In `from_dict`, when `strict=False`, filter `data` to known field names before calling `parse_class_config`:

```python
if strict:
    ...
    if extras:
        raise TypeError(...)
else:
    data = {k: v for k, v in data.items() if k in field_names}
return parse_class_config(cls, data)
```

### Environment

- physicalai (runtime) commit: `a7ee74d0a65bd8f27dc31d831210c7a905e92309`
- Downstream: physical-ai-studio issue open-edge-platform/physical-ai-studio#1069

Daankrol:
### What

`Config.from_dict(data, strict=False)` is documented to reject unknown keys only when `strict=True`, implying `strict=False` tolerates extra keys. It doesn't: `from_dict` still forwards the full, unfiltered `data` to `parse_class_config()`, which builds a jsonargparse parser from only the dataclass's known fields and unconditionally rejects any key without a matching argument. `strict=False` is currently a no-op — it can never succeed where `strict=True` also fails.

This PR filters `data` down to the dataclass's known top-level fields before delegating to `parse_class_config()` when `strict=False`.

### Why this is safe

This is a regression, not new behavior. The implementation before the jsonargparse-based rewrite (`dict_to_dataclass`) iterated only over the dataclass's own fields and silently ignored extras when `strict=False`. I verified this fix restores **exact parity** with that prior behavior, including a detail that's easy to get wrong: nested unknown keys (i.e. extras inside a field's own dict value) were **not** tolerated even by the old `strict=False`, and this fix preserves that — only top-level keys are filtered.

| | top-level extra key | nested extra key |
|---|---|---|
| pre-regression (`dict_to_dataclass`) | silently dropped | `TypeError` |
| current `main` | `ArgumentError` | `ArgumentError` |
| this PR | silently dropped | still rejected |

### Regression context

Bisected to the switch from `dict_to_dataclass` to the jsonargparse-based `parse_class_config` in the `Config.from_dict` rewrite. Notably, `tests/unit/config/test_config.py` has `test_typed_dataclass_strict_rejects_extra_keys` pinning `strict=True` behavior but had no equivalent test for `strict=False` — this PR adds one.

### Downstream impact

This broke `open-edge-platform/physical-ai-studio#1069`: loading any pretrained `lerobot`-format `config.json` (e.g. `lerobot/pi05_base`) via `SomeConfig.from_dict(hf_config, strict=False)` failed, because `lerobot` config files always carry keys with no matching field (`input_features`, `output_features`, `repo_id`, `license`, `push_to_hub`, ...). Studio shipped a downstream workaround (pre-filtering keys before calling `from_dict`) that can be reverted once this lands and is picked up in their runtime pin.

### Testing

- Added `test_typed_dataclass_non_strict_drops_unknown_top_level_keys` (the missing regression test) and `test_typed_dataclass_non_strict_still_rejects_nested_unknown_keys` (pins the nested-parity behavior) to `tests/unit/config/test_config.py`.
- `uv run pytest tests/unit/config/ -q` → 121 passed.
- `uv run pytest tests/unit -q -m "not slow and not requires_download and not integration and not v4l2 and not ov_smoke"` → same pass/fail set as `main` (6 pre-existing failures, all `ModuleNotFoundError: msgpack`, confirmed present on `main` too and unrelated to this change).
- `uv run ruff check` / `ruff format --diff` on changed files → no new findings vs. `main`.
- `uv run mypy src/physicalai/config/base.py` → same single pre-existing finding as `main` (unrelated, `__hash__ = None` on line 100).

Closes #251

### DIFF
--- a/src/physicalai/config/base.py
+++ b/src/physicalai/config/base.py
@@ -182,7 +182,13 @@ class Config:
 
         Args:
             data: Recipe envelope or dataclass field mapping.
-            strict: When reconstructing a typed dataclass, reject unknown keys.
+            strict: When reconstructing a typed dataclass, reject unknown
+                top-level keys. When ``False``, keys with no matching field on
+                ``cls`` are silently dropped instead of being forwarded to the
+                underlying parser (which would otherwise reject them
+                regardless of ``strict``). This only applies at the top
+                level: unknown keys nested inside a field's own value are not
+                filtered and are still rejected, matching prior behavior.
 
         Returns:
             A :class:`Config` recipe or dataclass instance.
@@ -198,12 +204,14 @@ class Config:
         if not dataclasses.is_dataclass(cls):
             msg = f"{cls.__name__} must be a dataclass to use Config"
             raise TypeError(msg)
+        field_names = {field.name for field in dataclasses.fields(cls)}
         if strict:
-            field_names = {field.name for field in dataclasses.fields(cls)}
             extras = set(data) - field_names
             if extras:
                 msg = f"Unexpected keys for {cls.__name__}: {sorted(extras)}"
                 raise TypeError(msg)
+        else:
+            data = {key: value for key, value in data.items() if key in field_names}
         return parse_class_config(cls, data)
 
     def instantiate(self, *, expected_type: type[_T] | None = None) -> _T | object:
--- a/src/physicalai/inference/postprocessors/stats_denormalizer.py
+++ b/src/physicalai/inference/postprocessors/stats_denormalizer.py
@@ -71,7 +71,7 @@ class StatsDenormalizer(Postprocessor):
         features: list[str] | None = None,
         *,
         artifact: str | None = None,
-        stats: dict[str, dict[str, np.ndarray]] | None = None,
+        stats: dict[str, dict[str, Any]] | None = None,
     ) -> None:
         """Initialize with stats file path and denormalization config.
 
@@ -85,7 +85,8 @@ class StatsDenormalizer(Postprocessor):
             artifact: Alias for *stats_path*, used when the path is
                 supplied via manifest ``artifact`` resolution.
             stats: Optional pre-loaded stats dict.  If provided, skips
-                lazy loading from file.
+                lazy loading from file.  Leaf values may be lists, scalars,
+                or numpy arrays; they are coerced to ``np.ndarray``.
 
         Raises:
             ValueError: If *mode* is not a recognized normalization mode
--- a/src/physicalai/inference/preprocessors/stats_normalizer.py
+++ b/src/physicalai/inference/preprocessors/stats_normalizer.py
@@ -70,7 +70,7 @@ class StatsNormalizer(Preprocessor):
         features: list[str] | None = None,
         *,
         artifact: str | None = None,
-        stats: dict[str, dict[str, np.ndarray]] | None = None,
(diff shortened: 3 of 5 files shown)


alfieroddan (dismissed):
Nice, literally just had this issue as well :smile:

maxxgx (approved):

PR #261, opened 2026-09-10
feat: integrate studio plugin package

MarkRedeman:
Move the [physicalai-studio-plugin](https://github.com/open-edge-platform/physical-ai-studio/tree/main/application/plugin) package from studio to the physicalai repo.
The goal here is that we will also move some of our plugins to this repo. Having the studio plugin here makes it easier to maintain as we can immediately change plugins when making changes to the plugin interface, and avoid cyclic dependency issues.

I've updated the release please configuration so that it makes a separate PR for each package, this way we can release them independently of the physicalai library.


MarkRedeman:
Closing so that I can create this as a stacked PR

PR #262, opened 2026-09-10
feat: integrate studio plugin package

MarkRedeman:
Move the [physicalai-studio-plugin](https://github.com/open-edge-platform/physical-ai-studio/tree/main/application/plugin) package from studio to the physicalai repo.
The goal here is that we will also move some of our plugins to this repo. Having the studio plugin here makes it easier to maintain as we can immediately change plugins when making changes to the plugin interface, and avoid cyclic dependency issues.

I've updated the release please configuration so that it makes a separate PR for each package, this way we can release them independently of the physicalai library.


.github/workflows/publish.yml
```diff
@@ -27,94 +37,123 @@ jobs:
...
-        (needs.release-please.result == 'success' && needs.release-please.outputs.release_created == 'true') ||
-        needs.release-please.result == 'skipped'
-      )
+    strategy:
+      matrix:
+        package: ${{ fromJSON(needs.release-please.outputs.packages) }}
     permissions:
       contents: read
+    env:
+      RELEASE_CREATED: ${{ fromJSON(needs.release-please.outputs.rp_outputs)[format('{0}--release_created', matrix.package.path)] }}
```
AlexanderBarabanov, line 49: As far as I understand, `release-please-action` prefixes outputs with `<path>--` for **non-root** paths, for the root package (path == ".") it emits unprefixed release_created/tag_name and therefore for `physicalai`  the workflow always looks up `.--release_created` which never exists ->  `physicalai` will never be published.

Please consider to add "normalization" step after `Run release-please` step to cover this case.

.github/release-please-config.json
```diff
@@ -22,6 +23,12 @@
     ".": {
```
AlexanderBarabanov, line 23: Please consider to add 
` "exclude-paths": ["packages/physicalai-studio-plugin"]` 
here, so that PR into just `packages/physicalai-studio-plugin` will not trigger release PR update for `physicalai`

https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md?plain=1#L309-L313

MarkRedeman: Thanks! Updated.

.github/workflows/publish.yml
MarkRedeman: Ah yeah that makes sense, I initially tried to build this based on my plugins repo but that one didn't use a root package.
In the recent commit I had OpenCode update the config based on your comment. 
It took me a while to understand it but I believe it works..

For reference here's the explanation that I asked it for: [release-please-config-step-by-step.md](https://github.com/user-attachments/files/32182920/release-please-config-step-by-step.md)

I feel like it's quite complicated, but it does make sense to me.

AlexanderBarabanov (approved):
Thank you, Mark!

MarkRedeman:
Merging this now, will do the following follow up steps:
1. After this gets merged merge the release-please pr so that we release a `0.2.0` version of the plugin package
2. Update all plugins so they depend on this latest version
3. Remove the `physicalai-studio-plugin` code from [studio](https://github.com/open-edge-platform/physical-ai-studio/tree/main/application/plugin), install from pypi instead
4. Merge the other stacked prs

PR #282, opened 2026-09-16
fix(rldx1): add runtime image resizing with stage-3 geometry controls

eugene123tw:
## Summary

- Added runtime image resizing in the RLDX-1 preprocessor so input frames are normalized to the exported image resolution before tokenization.
- Introduced optional Stage-3 geometry parameters for preprocessing: image_max_area, image_min_area, and image_resize_m.
- Implemented a robust resize pipeline with intermediate shape computation, center crop/pad, and contiguous output handling.
- Expanded unit tests to cover:
  - Backward-compatible resizing for older manifests that only provide image_resolution.
  - Stage-3 area-budget resizing behavior using manifest-provided geometry parameters.
- Improved preprocessor docstrings to clarify behavior and compatibility paths.

## Why

- Runtime can receive raw camera resolutions that do not match the frozen export resolution, which can cause geometry mismatches in downstream vision tokenization.
- Newer exports may carry Stage-3 area-budget resize settings, and runtime should honor those settings for parity with training/export preprocessing.
- Backward compatibility is required so older manifests continue to work without requiring re-export.

## Validation

- Ran targeted unit tests:
  - uv run pytest `test_rldx1.py`
- Result:
  - 7 passed in 0.31s

## Breaking changes

- None

## Related issues

- Closes #


samet-akcay:
Same as this comment https://github.com/openvinotoolkit/physicalai/pull/234#issuecomment-5692914968

can we have this following structure for better organization

```text
  src/physicalai/inference/preprocessors/
  ├── __init__.py
  ├── base.py
  ├── stats_normalizer.py
  ├── resize.py
  ├── pi05.py
  ├── rldx1/
  │   ├── __init__.py
  │   ├── processor.py
  │   ├── rope.py
  │   └── token_composer.py
  └── molmoact2/
      ├── __init__.py
      ├── processor.py
      ├── inputs.py
      └── image.py
```

we could have relative imports in init, so the existing imports would still work ?

samet-akcay (approved):

Issue #220, opened 2026-08-05
Pass the image layout directly instead of using an heuristic


AlexanderBarabanov:
Within this task please check if we could pass the image layout directly instead of using an heuristic in
https://github.com/AlexanderBarabanov/physicalai/blob/main/src/physicalai/inference/preprocessors/resize.py
https://github.com/AlexanderBarabanov/physicalai/blob/main/src/physicalai/inference/preprocessors/smolvla.py

 See this PR https://github.com/openvinotoolkit/physicalai/pull/217

johyeongseob:
Hi @AlexanderBarabanov , I'd like to work on this issue. 

I'll review the current layout detection in resize.py and smolvla.py and the changes from #217, then investigate how we can pass the image layout explicitly instead of relying on the heuristic. 

Please let me know if it's okay for me to take this. 

Thanks!

AlexanderBarabanov:
Hi @johyeongseob , Thank you for the proposal! I've assigned this task to you.
fyi: @maxxgx @sovrasov

johyeongseob:
@AlexanderBarabanov 

Thanks for assigning this to me! 

I'll start looking into the implementation and will open a PR once I have a working solution.

johyeongseob:
Hi @AlexanderBarabanov ,

These checks were run on the implementation in PR #285 (commit `3c6b166`).

I ran additional manual checks using synthetic images on Ubuntu 24.04 (WSL), with Python 3.12.3.

Both BCHW and BHWC inputs were checked against explicitly constructed expected outputs:

1. **Ambiguous shape `(1, 3, 3, 3)`**: distinct pixel values verify that both preprocessors interpret the specified layout correctly.
2. **Resizing a constant-color 2×4 image to 6×8**: verifies stretch, centered letterbox padding, and SmolVLA's top padding, normalization, and camera masks.

All assertions passed. The constant-color case checks resize geometry and padding; it does not exhaustively test interpolation.

Reproduction code (`validation.py`):

```python
import numpy as np
from physicalai.inference.preprocessors import ResizePreprocessor, ResizeSmolVLA

def check(image, size, mode, expected, smol=False):
    for layout in ("BCHW", "BHWC"):
        pixels = image if layout == "BCHW" else image.transpose(0, 2, 3, 1)
        args = dict(image_resolution=size, image_layout=layout)
        prep = ResizeSmolVLA(**args) if smol else ResizePreprocessor(**args, mode=mode)
[...]

PR #285, opened 2026-09-17
fix(inference): replace image layout heuristics with explicit configuration

johyeongseob:
## Summary

- Add a required keyword-only `image_layout` argument to `ResizePreprocessor` and `ResizeSmolVLA`, accepting `"BCHW"`, `"BHWC"`, or the corresponding `ImageLayout` enum values.
- Replace shape-based layout detection with explicit layout handling while preserving output layouts.
- Add tests for ambiguous shapes, equivalent pixels across layouts, camera masks, invalid or missing layouts, and configuration through both `type` and `class_path`.
- Update fuzz harnesses, manifest seeds, and configuration documentation.

## Why

Image dimensions alone cannot reliably determine the channel axis. For example, `(1, 3, 3, 3)` can represent either BCHW or BHWC input.

Passing the layout explicitly lets the caller specify the axis order without relying on channel-count heuristics.

## Validation

Environment:
- Ubuntu 24.04 on WSL
- Python 3.12.3
- Dependencies managed with uv

Results:
- `uv run --no-sync pytest tests/unit/inference/ -q`
  - 549 passed, 2 skipped, 2 warnings.
  - Skips: missing optional `torch` dependency and restricted PaliGemma tokenizer access.
  - Warnings: pytest collection of `TestAdapter` and multiprocessing `fork()` in the OpenVINO tokenizer test.
- Ruff 0.15.12 lint and format checks passed for the four changed preprocessor source files.
- Pyrefly 1.2.0 reported 112 errors on both the baseline (`de49e1f`) and this branch using the same Python environment. Sorted error messages and their occurrence counts matched.
  - Three targeted suppressions were added in tests: two intentional missing-argument calls and one nested-camera input supported at runtime but absent from the existing input annotation.

## Breaking changes

- Existing Python calls and configurations for these two preprocessors must now provide `image_layout`.
- For `type` configurations, add `image_layout` alongside the other arguments. For `class_path` configurations, add it under `init_args`.
- All images passed to a preprocessor instance must use the configured layout.
- Automatic layout detection is removed. Output layouts remain unchanged.

## Related issues

- Closes issue: #220
- Related PR: #217


src/physicalai/inference/preprocessors/resize.py
```diff
@@ -37,12 +40,14 @@ def __init__(
         self,
         image_resolution: tuple[int, int],
         *,
+        image_layout: ImageLayout | str,
```
sovrasov: To preserve compatibility with already released versions and previously trained models, this argument should be optional. In case of missing layout, auto-detection is performed in runtime.
I afraid, we'd have to retain old heuristics for that case, although they can be extracted in helper methods or functions instead of bloating the useful code.

sovrasov (changes requested):
@johyeongseob thanks for a nice PR! 
One comment from my side is that we have to preserve backward compatibility this time.

src/physicalai/inference/preprocessors/resize.py
```diff
@@ -124,15 +130,7 @@ def _resize_with_ar_pad(self, img: np.ndarray) -> np.ndarray:  # noqa: PLR0912,
...
-        # Heuristic: standard channel counts are {1, 2, 3, 4}; spatial dims are typically larger.
-        if img.shape[-1] in {1, 2, 3, 4} and img.shape[1] in {1, 2, 3, 4}:
-            msg = (
-                f"ambiguous layout: both dim 1 ({img.shape[1]}) and dim -1 ({img.shape[-1]}) "
-                "look like standard channel counts; provide input with spatial dims > 4"
-            )
-            raise ValueError(msg)
-        channels_last = img.shape[-1] in {1, 2, 3, 4} and img.shape[1] not in {1, 2, 3, 4}
-        if not channels_last:
+        if self._image_layout == ImageLayout.BCHW:
```
maxxgx: Could we add a check for the channel count here? If someone sets BCHW for a BHWC camera image, we can silently treat the image height as the number of channels. If we only support 1–4 channels, a clear error would help catch this. Same for ResizeSmolVLA.

src/physicalai/inference/preprocessors/resize.py
johyeongseob: Thank you for the feedback. I'll make image_layout optional and preserve the existing auto-detection for backward compatibility!

src/physicalai/inference/preprocessors/resize.py
johyeongseob: Thank you for the suggestion. As part of this update, I'll check the supported channel counts and add appropriate validation to both preprocessors!

johyeongseob:
Thank you for the feedback, @sovrasov . Addressed in b106c6c: `image_layout` is now optional, preserving the existing auto-detection when omitted or set to `None`.

I'll also follow up on @maxxgx 's suggestion to validate channel counts in both preprocessors.

johyeongseob:
Thank you for the suggestion, @maxxgx . 

Addressed in a846b57: both preprocessors now validate 1–4 channels when image_layout is explicitly specified, while preserving legacy behavior when omitted or set to None.

johyeongseob:
Thank you both for your helpful feedback, @sovrasov  and @maxxgx . 

I've addressed both review comments and would appreciate your review when you have a chance.

maxxgx (approved):
Thanks for the updates! Looks good to me

sovrasov (approved):

